PR #55, opened 2023-04-19
[QA-142] Shutdown OpenTelemetry collector at end of test

Tom-Dann:
##  QA-142

Adding a `kill` command in the post-build to send a `SIGTERM` signal to the OpenTelemetry collector, and wait for it to shutdown gracefully with a while loop

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -253,6 +253,8 @@ Resources:
                 - k6 run $WORK_DIR/$TEST_SCRIPT --tag script=$TEST_SCRIPT --tag account_id=${AWS::AccountId} --out statsd
             post_build:
               commands:
+                - echo Wait 5 minutes to flush remaining metrics
+                - sleep 300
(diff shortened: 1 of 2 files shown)


abhinaysrivastavaperf (approved):

beca-galliano (approved):

deploy/template.yaml
```diff
@@ -253,6 +253,8 @@ Resources:
                 - k6 run $WORK_DIR/$TEST_SCRIPT --tag script=$TEST_SCRIPT --tag account_id=${AWS::AccountId} --out statsd
             post_build:
               commands:
+                - echo Wait 5 minutes to flush remaining metrics
```
andyloughran, line 256: Can we find a better way to do this please?

andyloughran (changes requested):
Shouldn't include time-based waits - let's put something in that will check if the buffer is flushed.

jamesmelville-gds (approved):

PR #77, opened 2023-05-22
[QA-184] Adding di-accounts-performance changes

Tom-Dann:
##  QA-184

Adding changes from `di-accounts-performance`


Tom-Dann:
Replaced with #78

PR #95, opened 2023-05-26
[QA-187] Add changes from di-ipv-cri-kiwi-performance

Tom-Dann:
## QA-187

- Adding changes from [di-ipv-cri-kiwi-performance](https://github.com/alphagov/di-ipv-cri-kiwi-performance)
- Renamed the F2F URL environment variable


deploy/template.yaml
```diff
@@ -1,7 +1,7 @@
 AWSTemplateFormatVersion: "2010-09-09"
 Transform: AWS::Serverless-2016-10-31
 Description: >
-  di-devplatform-performance
+  di-ipv-cri-kiwi-performance
```
abhinaysrivastavaperf, line 4: Do we need to update the description in the CF template?

deploy/template.yaml
```diff
@@ -269,7 +270,7 @@ Resources:
         - Key: "Service"
           Value: "ci/cd"
         - Key: "Source"
-          Value: "alphagov/di-devplatform-performance"
+          Value: "alphagov/di-ipv-cri-kiwi-performance"
```
abhinaysrivastavaperf, line 273: The value of the tag doesn't need to be updated for this change.

abhinaysrivastavaperf (changes requested):

deploy/template.yaml
Tom-Dann: Thanks - I'll revert these ones

abhinaysrivastavaperf (approved):

PR #123, opened 2023-06-09
[QA-41]: Updated workload model for Account Management scenarios

abhinaysrivastavaperf:
### What?
Updated the workload model of the account management scenarios

#### Changes:
- Target load -  10 iterations per second
- Ramp up duration - 15 minutes
- Steady state - 30 minutes

---

### Why?
This is the proposed workload model for initial performance tests for Accounts.

### DIFF
--- a/deploy/scripts/src/accounts/test.ts
+++ b/deploy/scripts/src/accounts/test.ts
@@ -63,10 +63,10 @@ const profiles: ProfileList = {
       startRate: 1,
       timeUnit: '1s',
       preAllocatedVUs: 1,
-      maxVUs: 300,
+      maxVUs: 1000,
       stages: [
-        { target: 10, duration: '30m' }, // Ramp up to 30 iterations per second in 30 minutes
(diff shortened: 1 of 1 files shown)


deploy/scripts/src/accounts/test.ts
```diff
@@ -63,10 +63,10 @@ const profiles: ProfileList = {
       startRate: 1,
       timeUnit: '1s',
       preAllocatedVUs: 1,
-      maxVUs: 300,
+      maxVUs: 1000,
```
Tom-Dann, line 66:
```suggestion
      maxVUs: 300,
```
This should be kept at `300`.

Max expected journey duration is `30` seconds
Target load is `10 `requests per second target
∴ Max VUs should be `30 * 10 = 300`

abhinaysrivastavaperf: Updated in the next commit

Tom-Dann (approved):

PR #133, opened 2023-06-15
[QA-193] Updated the CIC journey

gpopescu23:
### What?
Updated the CIC journey with a POST/checkDetails

#### Changes:

- `cri-kiwi/test.ts`
   - Changed the first request with a POST/start for calling the stub
   - Added POST/checkDetails where the urlBody will be checked if contain the clientId
  
- `deploy/template.yaml`
   - Added the new environment variables

---

### Why?
The BE has been integrated with the FE. 

---

[QA-193]: https://govukverify.atlassian.net/browse/QA-193?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


deploy/scripts/src/cri-kiwi/test.ts
```diff
@@ -78,19 +78,47 @@ export function setup (): void {
...
+    })
+      ? transactionDuration.add(endTime - startTime)
+      : fail('Response Validation Failed')
+    requestValue = getRequestCode(res)
+    clientId = getClientID(res)
+  })
+
+  group('B01_CIC_02_Authorize GET', function () {
+    const startTime = Date.now()
+    res = http.get(env.envURL + '/oauth2/authorize?request=' + requestValue + '&response_type=code&client_id=' + clientId, {
```
Tom-Dann, line 120:
```suggestion
    res = http.get(env.envURL + '/oauth2/authorize',
    {
      request: requestValue,
      response_type: code,
      client_id: clientId
    },
    {
```
This would be more readable in `object` format

Tom-Dann (changes requested):

deploy/scripts/src/cri-kiwi/test.ts
```diff
@@ -78,19 +79,45 @@ export function setup (): void {
...
 }
 
 const transactionDuration = new Trend('duration')
 
 export function CIC (): void {
   let res: Response
   let csrfToken: string
+  let requestValue: string
+  let clientId: string
+  const encodeRequest = b64encode('{"alg":"RSA', 'rawstd')
```
Tom-Dann, line 94: This doesn't need to be a variable if we are just using it in one place

deploy/scripts/src/cri-kiwi/test.ts
```diff
@@ -717,3 +761,37 @@ function randomDate (start: Date, end: Date): Date {
...
+    if (response?.clientId != null) {
+      const clientId = response.clientId
+      return clientId
+    } else {
+      return 'No value found'
+    }
+  } else {
+    return 'Invalid response body'
+  }
+}
```
Tom-Dann, line 797:
```suggestion
function getClientID (r: Response): string {
  const clientId = r.json('clientId')
  if (clientId !== null && typeof clientId === 'string') return clientId
  fail('Client ID not found')
}
```
This can be simplified using `Response.json()` built in function. Also we should call `fail()` in the case of unrecoverable errors, returning string values will cause the script to continue on.

Tom-Dann (changes requested):

Tom-Dann (approved):

PR #135, opened 2023-06-16
[QA-212] Split BuildSpec into its own file

Tom-Dann:
### What?
Split out the BuildSpec specifcation into its own file

#### Changes:
- Removed BuildSpec from `deploy/template.yml`
- Created a new file `deploy/buildspec.yml` which contains the information removed from the `template.yml`

---

### Why?
This change will make it simpler to manage the CodeBuild BuildSpec and changes in the future. Changes no longer need to be reviewed as a multi-line string in a YAML file, and can be modified as a YAML file in of itself.

---

### Related:
- [AWS CloudFormation Documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-codebuild-project-source.html#cfn-codebuild-project-source-buildspec)
- [AWS CodeBuild BuildSpec Documentation](https://docs.aws.amazon.com/codebuild/latest/userguide/build-spec-ref.html)


deploy/template.yaml
```diff
@@ -197,77 +201,6 @@ Resources:
       TimeoutInMinutes: 480
       Source:
         Type: "NO_SOURCE"
```
andyloughran, line 203: How does the BuildSpec work now if it's no longer referenced here?

andyloughran: The build specification for the project. If this value is not provided, then the source code must contain a buildspec file named buildspec.yml at the root level. If this value is provided, it can be either a single string containing the entire build specification, or the path to an alternate buildspec file relative to the value of the built-in environment variable CODEBUILD_SRC_DIR. The alternate buildspec file can have a name other than buildspec.yml, for example myspec.yml or build_spec_qa.yml or similar. For more information, see the [Build Spec Reference](https://docs.aws.amazon.com/codebuild/latest/userguide/build-spec-ref.html#build-spec-ref-example) in the AWS CodeBuild User Guide.

andyloughran: @Tom-Dann can you put together a doc on how this buildspec will be governed, and what parameters need to be in place to ensure they don't do things they shouldn't - plus do the things that are required (e.g. mk test report .etc) - or just to document that teams should follow and copy/paste your example, then how would they keep checking back in for updates?

Tom-Dann: The BuildSpec governance remains the same, it is just now in it's own file which helps with linting it as native YAML rather than a multi-line string in the context of the larger CloudFormation template

PR #149, opened 2023-06-22
[QA-218] Store JSON test results in S3

Tom-Dann:
### What?
Test results file in JSON format is now saved to an S3 bucket in the account

> **Note**
> This change was deployed and tested in `development`
> 
> image

#### Changes:
- `deploy/template.yml`
  - Added a new S3 bucket to store test results in
  - Added `s3:PutObject` policy to the CodeBuild service policy to give permissions to the CodeBuild agent to copy files to the S3 bucket
  - Moved some Environment Variables defined in the CodeBuild spec: some of these are needed for the S3 bucket location and some are moved so that they are more easily editable when running `Start Build with Overrides`
  - Added an additional `--out` flag to the `k6` command to also output results in JSON format and zipped as a `.gz` file

---

### Why?
For posterity and as a backup it is needed to persist flat files of test results

---

### Related:
- [AWS CloudFormation S3 Bucket Documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket.html)
- [k6 JSON Output Documentation](https://k6.io/docs/results-output/real-time/json/)


deploy/template.yaml
```diff
@@ -174,6 +179,29 @@ Resources:
...
                             Fn::Sub: "${VpcStackName}-ProtectedSubnetIdB"
 
+  ResultsBucket:
+    Type: AWS::S3::Bucket
+    Properties:
+      AccessControl: Private
+      BucketEncryption:
+        ServerSideEncryptionConfiguration:
+          - ServerSideEncryptionByDefault:
+              SSEAlgorithm: AES256
```
andyloughran, line 189: What data is going into this bucket?
Should it be controlled with a KMS key instead to have tighter control of the data and what can decrypt/encrypt it?

deploy/template.yaml
```diff
@@ -174,6 +179,29 @@ Resources:
...
+      BucketName: !Sub ${AWS::StackName}-test-results
+      LifecycleConfiguration:
+        Rules:
+          - ExpirationInDays: 1825 # Five years
```
andyloughran, line 193: This feels like a very long time, can you check the policy with Ben/Security?

deploy/template.yaml
Tom-Dann: It is for storage of test result flat files

deploy/template.yaml
Tom-Dann: Reduced to 365 days

PR #159, opened 2023-06-27
DCMAW-5594: Add BE e2e test script

jamielutton-gds:
### What?

Added test script for performance test of BE e2e journey:
- POST Test client /start
- POST /verifyAuthorizeRequest
- POST /resourceOwner/documentGroups
- GET /biometricToken/v2
- POST /finishBiometricSession
- GET /redirect
- POST /token
- POST /userinfo/v2

#### Changes:
- Created test script within `/mobile/backend-e2e.test.ts`
- Created functions for making BE API calls (`postVerifyAuthorizeRequest`, `postResourceOwnerDocumentGroups`, `getBiometricTokenV2`, `postFinishBiometricSession`, `getRedirect`, `postToken`, `postUserInfoV2`)
- Updated FE e2e script to call new `getBiometricTokenV2` and `postFinishBiometricSession` functions to finish session
- Broke down functions in `mobile/utils/functions-mobile-journey.ts` into separate files for better separation of concerns: `assertions.ts`, `backend.ts`, `config.ts`, `frontend.ts`, `test-client.ts`, `url.ts`)

---

### Why?

Creation of BE test script to test load on system.
File restructure to reduce complexity of utils function file.

---

### Related:
- [Ticket](https://govukverify.atlassian.net/browse/DCMAW-5594)


deploy/scripts/src/mobile/utils/test-client.ts
```diff
@@ -0,0 +1,23 @@
+import http, { type Response } from 'k6/http'
+import { getUrl } from './url'
+import { config } from './config'
+
+export function postTestClientStart (): Response {
+  return http.post(
+    getUrl('start', config.testClientExecuteUrl),
+    JSON.stringify({ frontendUri: config.frontendUrl }),
```
jamielutton-gds, line 8: Note: We have removed `target` from the request body here due to issues in the test client configuration which we've discussed before. We ran the tests locally against build to ensure there were no issues with running it against that environment without specifying the target here.

deploy/scripts/src/mobile/backend-e2e.test.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+}
+
+const loadProfile = selectProfile(profiles)
+
+export const options: Options = {
+  // httpDebug: 'full',
+  scenarios: loadProfile.scenarios,
+  thresholds: {
+    http_req_duration: ['p(95)<=1000', 'p(99)<=2500'], // 95th percentile response time <=1000ms, 99th percentile response time <=2500ms
```
jamielutton-gds, line 37: Is this something people feel strongly about including? We kept it as it was left over from the FE journey and seems like reasonable expectations of API calls. All it seems to do is add a ✅  or ❌  next to the test output if it fails, and logs a message if the threshold isn't met. 

The POST /userinfo/v2 call frequently causes the 95th percentile to exceed 1000ms anyway, so not sure what value this really adds other than flagging that. 
image

sandy-may: I feel strongly about it! It's automatic threshold validation that will cause the test to fail if criteria aren't met.

It checks these two NFRs @jamielutton-gds 

image

sandy-may: I'm not surprised we see it failing when running smoke tests - we'll hit lambda cold starts for both userInfo and the readId mock.

jamielutton-gds: Thanks for the context! What does this mean then - do our NFRs have to change to account for this? Or in the future, will we have finer-grained tests, that would check (as an example) something like p(95) < 1s for all endpoints except userinfo, but p(95)<1.5 for userinfo only?

sandy-may: We'll need to run this performance test at load to get a full picture of how userInfo really performs

sandy-may: These NFRs are set at a programme level and if we can't meet them we have a defect to triage/fix

jamielutton-gds: That makes sense. Thanks!

deploy/scripts/src/mobile/README.md
```diff
@@ -23,8 +23,8 @@ npm install
 
 
 ## Run Command
-K6 environment variales can be found in test-mobile-journey.ts file.
-Profile and scenario can be found in test.ts file.
+K6 environment variales can be found in config.ts file.
```
sandy-may, line 26: whilst we're here, please may we fix the spelling of variables

deploy/scripts/src/mobile/utils/backend.ts
sandy-may, line 1: Would it make sense to have a folder under src/mobile called testSteps?

This file feels much more than a utility to me

deploy/scripts/src/mobile/utils/frontend.ts
sandy-may, line 1: Similar here - this feels much more important than a utility

sandy-may: (not that utilities aren't important...)

deploy/scripts/src/mobile/utils/config.ts
```diff
@@ -0,0 +1,5 @@
+export const config = {
```
sandy-may, line 1: Would it help to add comments indicating what the __ENV means?

deploy/scripts/src/mobile/utils/url.ts
```diff
@@ -0,0 +1,28 @@
...
+  if (query != null) {
+    Object.entries(query).forEach(([key, value]) => {
+      url.searchParams.set(key, value)
+    })
+  }
+
+  return url.toString()
+}
+
+export function getFrontendUrl (path: string, query?: Record<string, string>): string {
```
sandy-may, line 22: could we name this to something more specific? It sounds to me like it's fetching the config.frontendUrl, but it's actually building a request

deploy/scripts/src/mobile/utils/assertions.ts
```diff
@@ -0,0 +1,54 @@
...
+  })
+}
+
+export function validatePageContent (res: Response, pageContent: string): boolean {
+  return check(res, {
+    'validate page content': (r) => (r.body as string).includes(pageContent)
+  })
+}
+
+export function validateHeaderLocation (res: Response): boolean {
```
sandy-may, line 38: minor - would it make sense for this to be validateLocationHeader?

deploy/scripts/src/mobile/backend-e2e.test.ts
```diff
@@ -0,0 +1,60 @@
...
+      ],
+      exec: 'backendJourney'
+    }
+  }
+}
+
+const loadProfile = selectProfile(profiles)
+
+export const options: Options = {
+  // httpDebug: 'full',
```
sandy-may, line 34: do we need this comment?

deploy/scripts/src/mobile/utils/backend.ts
```diff
@@ -0,0 +1,106 @@
...
+    const res = http.post(documentGroupsUrl, JSON.stringify(documentGroupsData))
+    isStatusCode200(res)
+  })
+}
+
+export function getBiometricTokenV2 (sessionId: string): void {
+  group('GET /biometricToken/v2', () => {
+    const biometricTokenUrl = getBackendUrl('/biometricToken/v2', { authSessionId: sessionId })
+    const res = http.get(biometricTokenUrl, {
+      tags: { name: 'Get Biometric Token' }
```
sandy-may, line 45: Are we expecting a tag here? We don't seem to be using them in other requests

deploy/scripts/src/mobile/utils/backend.ts
```diff
@@ -0,0 +1,106 @@
...
+interface RedirectResponse {
+  authorizationCode: string
+  redirectUri: string
+}
+
+export function getRedirect (sessionId: string): RedirectResponse {
+  return group('GET /redirect', () => {
+    const redirectUrl = getBackendUrl('/redirect', { sessionId })
+    const redirectRes = http.get(redirectUrl, {
+      tags: { name: 'Redirect BE' }
```
sandy-may, line 71: same here - do we want tags?

deploy/scripts/src/mobile/utils/backend.ts
```diff
@@ -0,0 +1,106 @@
...
+    return {
+      authorizationCode: redirectRes.json('authorizationCode') as string,
+      redirectUri: redirectRes.json('redirectUri') as string
+    }
+  })
+}
+
+export function postToken (redirectResponse: RedirectResponse): string {
+  return group('POST /token', () => {
+    const accessTokenUrl = getBackendUrl('/token')
```
sandy-may, line 84: nit: the api here is just token - should we update the variable names to reflect that? i.e. remove access

deploy/scripts/src/mobile/utils/backend.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    isStatusCode200(redirectRes)
+    return {
+      authorizationCode: redirectRes.json('authorizationCode') as string,
+      redirectUri: redirectRes.json('redirectUri') as string
+    }
+  })
+}
+
+export function postToken (redirectResponse: RedirectResponse): string {
```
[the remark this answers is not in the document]
jamielutton-gds: It felt a bit off to me too... Maybe `redirectData` would be better?

deploy/scripts/src/mobile/utils/test-client.ts
sandy-may: not quite in answer to your point, but what's the rationale for needing a frontendUri? I thought that was only needed when using the default test client with your own FE stack

jamielutton-gds: I think if you omit frontendUri then the WebLocation will use the AWS resource ID instead of the pretty URL. Seems better to use the pretty URL if we can, what do you think? (Hence why I was saying we would have provided the `target` as well, had it not been for the issue with the test client)

deploy/scripts/src/mobile/README.md
jamielutton-gds: Done

deploy/scripts/src/mobile/utils/frontend.ts
jamielutton-gds: Done

deploy/scripts/src/mobile/utils/test-client.ts
sandy-may: ah that makes sense - we should really look at updating the test client!

deploy/scripts/src/mobile/utils/config.ts
jamielutton-gds: Have added a comment here - do you think this makes it clearer?

deploy/scripts/src/mobile/utils/url.ts
jamielutton-gds: Updated these function names to `buildUrl`, `buildFrontendUrl` and `buildBackendUrl`

deploy/scripts/src/mobile/utils/assertions.ts
jamielutton-gds: Yep. Done!

deploy/scripts/src/mobile/utils/backend.ts
jamielutton-gds: I think this was left over from us copying across some FE stuff. Since we've discussed already not adding things where we don't know what it does and without knowing the value of it, I've removed it

deploy/scripts/src/mobile/utils/backend.ts
jamielutton-gds: Removed, same as above

deploy/scripts/src/mobile/utils/backend.ts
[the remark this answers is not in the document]
jamielutton-gds: We weren't sure what people would think about this, we did consider both options - given you feel this is fine, updated the signature here!

deploy/scripts/src/mobile/utils/backend.ts
jamielutton-gds: Good catch, fixed

deploy/scripts/src/mobile/backend-e2e.test.ts
jamielutton-gds: Again, I think this was left over from the FE config we copied across. With this option enabled, it logs the entire requests and responses in full, but is this something that would be useful for us in debugging? It logs a lot of information, even the HTML returned from the GET requests, so thought it was maybe providing too much detail

@raluca-alecu Tagging you in case you have any further insight on the usefulness of this option!

jamielutton-gds: None of the other performance test scripts elsewhere in the repo have this option enabled. I'd be happy to omit, given we generally don't leave comments unless they're necessary to explain something

jamielutton-gds: I think the logs are also a bit hard to digest at the level of detail that they provide, if you have lots of iterations running at once I feel it wouldn't be easy to sift through them (they also get truncated by my IDE terminal, there are that many of them 😅 )

sandy-may: Yep we definitely don't want this option enabled when running a load test - it will slow the test tool down. 

I'll leave it up to you whether to leave it commented out or remove from the file entirely

deploy/scripts/src/mobile/utils/url.ts
sandy-may: nice - looks good

deploy/scripts/src/mobile/backend-e2e.test.ts
raluca-alecu: I think we can remove the comment 👍

deploy/scripts/src/mobile/utils/config.ts
sandy-may: I was thinking of just saying something __ENV is the k6 shorthand for accessing environment variables. See here deploy/scripts/README.md for guidance on setting these variables?

sandy-may (approved):

deploy/scripts/src/mobile/backend-e2e.test.ts
jamielutton-gds: Removed!

deploy/scripts/src/mobile/utils/config.ts
jamielutton-gds: Didn't notice it was already well-documented there. Have updated!

deploy/scripts/src/mobile/README.md
```diff
@@ -34,5 +34,4 @@ npm start
 
 ```bash
 # Run k6 test. Replace with actual values
-k6 run ./dist/mobile/test.js -e PROFILE=<your_profile_name> -e SCENARIO=<your_scenario_name> -e MOBILE_TEST_CLIENT_EXECUTE_URL=<your_test_client_execute_url> -e MOBILE_BACK_END_URL=<your_backend_url> -e MOBILE_FRONT_END_URL=<your_frontend_url> -e MOBILE_BIOMETRIC_SESSION_ID=<your_biometric_session_id>
-```
\ No newline at end of file
+k6 run ./dist/mobile/<your_test_file>.test.js -e PROFILE=<your_profile_name> -e SCENARIO=<your_scenario_name> -e MOBILE_TEST_CLIENT_EXECUTE_URL=<your_test_client_execute_url> -e MOBILE_BACK_END_URL=<your_backend_url> -e MOBILE_FRONT_END_URL=<your_frontend_url> -e
```
Tom-Dann, line 37:
````suggestion
k6 run ./dist/mobile/<your_test_file>.test.js -e PROFILE=<your_profile_name> -e SCENARIO=<your_scenario_name> -e MOBILE_TEST_CLIENT_EXECUTE_URL=<your_test_client_execute_url> -e MOBILE_BACK_END_URL=<your_backend_url> -e MOBILE_FRONT_END_URL=<your_frontend_url>
```
````
Remove trailing `-e` flag and add back the closing ```` ``` ```` three backticks of the code block

PR #161, opened 2023-06-28
[QA-239] Make S3 bucket name globally unique

Tom-Dann:
### What?
Adding a portion of the AWS stack ID to end of the test results S3 bucket name

> **Note** 
> This was deployed to development and the s3 bucket with the new name was created
>
> image

#### Changes:
- Changed bucket name:
  - From: `${AWS::StackName}-test-results`
  - To: `${AWS::StackName}-test-results-<twelve_digit_stack_id>`

---

### Why?
Buckets deployed to production and non-production would end up with the same name, this adds a portion of the stack ID to the end of the bucket name for global uniqueness 

---

### Related:
- #149

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -187,7 +187,20 @@ Resources:
         ServerSideEncryptionConfiguration:
           - ServerSideEncryptionByDefault:
               SSEAlgorithm: AES256
-      BucketName: !Sub ${AWS::StackName}-test-results
+      BucketName:
+        Fn::Join:
+            - '-'
+            - - Ref: AWS::StackName
+              - test-results
+              - Fn::Select:
+                  - 4
+                  - Fn::Split:
+                      - '-'
+                      - Fn::Select:
+                          - 2
+                          - Fn::Split:
+                              - /
+                              - Ref: AWS::StackId
       LifecycleConfiguration:
         Rules:
           - ExpirationInDays: 365


deploy/template.yaml
```diff
@@ -187,7 +187,20 @@ Resources:
         ServerSideEncryptionConfiguration:
           - ServerSideEncryptionByDefault:
               SSEAlgorithm: AES256
-      BucketName: !Sub ${AWS::StackName}-test-results
+      BucketName:
+        Fn::Join:
+            - '-'
+            - - Ref: AWS::StackName
```
devdutoitGDS, line 193: does Ref work without the !

devdutoitGDS (approved):

deploy/template.yaml
Tom-Dann: Yes you can do `Ref:` or `!Ref`

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-ref.html#YAML

deploy/template.yaml
```diff
@@ -187,7 +187,20 @@ Resources:
...
+              - test-results
+              - Fn::Select:
```
jamielutton-gds, line 195: Do we need the select here? Can we not just use the entire suffix from the stackId?

deploy/template.yaml
```diff
@@ -187,7 +187,20 @@ Resources:
...
+                  - 4
+                  - Fn::Split:
+                      - '-'
+                      - Fn::Select:
+                          - 2
+                          - Fn::Split:
+                              - /
+                              - Ref: AWS::StackId
```
jamielutton-gds, line 203: Also, does the StackId change even on an update only? Instinctively, I would have thought that when a stack is created, it maintains the same StackId throughout its lifecycle (though I'm struggling to find confirmation of how it works in the docs!)

deploy/template.yaml
Tom-Dann: The 12 character end of the Stack UUID is sufficient to give uniqueness of the bucket name. It is also inline with where we have used the same 12 char ID in the suffix of the `VPCTestPolicy` name

deploy/template.yaml
Tom-Dann: The stacks are the promotion stacks, which remain the same. The non-prod and prod stacks have different IDs though

jamielutton-gds: Ah, sorry, I realise I just misunderstood the issue here.

deploy/template.yaml
jamielutton-gds: Ah, I didn't realise that the last 12 characters were sufficient to guarantee uniqueness - if that's the case, I'm happy for this to be resolved. (For my benefit though, is there a link to documentation which says this? 😅 )

Tom-Dann: There is some info on the format of the Stack ID [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/pseudo-parameter-reference.html#cfn-pseudo-param-stackid). It's the end of the UUID we are using here

PR #164, opened 2023-06-29
DCMAW-5611: Smoke test journeys in Build

raluca-alecu:
### What?
Added BE URL to the test client request

#### Changes:
- added backend URL to the postTestClientStart() function

---

### Why?
build is failing due to the allowed domains specified in the egress rules in the performance test account. Adding the BE URL to the test client request will return the ApiLocation with an allowed domain.

---

### DIFF
--- a/deploy/scripts/src/mobile/utils/test-client.ts
+++ b/deploy/scripts/src/mobile/utils/test-client.ts
@@ -5,7 +5,7 @@ import { config } from './config'
 export function postTestClientStart (): Response {
   return http.post(
     buildUrl('start', config.testClientExecuteUrl),
-    JSON.stringify({ backendUri: `${config.backendUrl}/`, frontendUri: config.frontendUrl }),
+    JSON.stringify({ target: config.backendUrl, frontendUri: config.frontendUrl }),
     {
       tags: { name: 'Post request to authorize URL' },
       headers: { 'Content-Type': 'application/json' }
@@ -19,5 +19,5 @@ export function parseTestClientResponse (response: Response, location: 'WebLocat
     throw new Error('Failed to parse URL from response')
   }
 
-  return url
+  return url.replace(config.backendUrl, `${config.backendUrl}/`)
 }
--- a/deploy/scripts/src/mobile/utils/test-client.ts
+++ b/deploy/scripts/src/mobile/utils/test-client.ts
@@ -5,7 +5,7 @@ import { config } from './config'
 export function postTestClientStart (): Response {
   return http.post(
     buildUrl('start', config.testClientExecuteUrl),
-    JSON.stringify({ frontendUri: config.frontendUrl }),
+    JSON.stringify({ backendUri: `${config.backendUrl}/`, frontendUri: config.frontendUrl }),
     {
       tags: { name: 'Post request to authorize URL' },
       headers: { 'Content-Type': 'application/json' }
--- a/deploy/scripts/src/mobile/utils/test-client.ts
+++ b/deploy/scripts/src/mobile/utils/test-client.ts
@@ -32,16 +32,17 @@ export function parseTestClientResponse (
 }
 
 function parseApiLocation (apiLocation: string): string {
-  const requestJwt = parseQueryParams(apiLocation, 'request')
-  const clientId = parseQueryParams(apiLocation, 'client_id')
+  function parseQueryParams (queryParams: string[], parameterName: string): string {
+    return queryParams.filter((value) => value.startsWith(parameterName))[0].split('=')[1]
+  }
+
+  const queryParams = apiLocation.split('?')[1].split('&')
+  const requestJwt = parseQueryParams(queryParams, 'request')
+  const clientId = parseQueryParams(queryParams, 'client_id')
 
   return buildBackendUrl('verifyAuthorizeRequest', {
     client_id: clientId,
     response_type: 'code',
     request: requestJwt
   })
-}
-
-function parseQueryParams (location: string, parameterName: string): string {
-  return location.split('?')[1].split('&').filter((value) => value.startsWith(parameterName))[0].split('=')[1]
 }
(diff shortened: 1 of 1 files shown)


deploy/scripts/src/mobile/utils/test-client.ts
```diff
@@ -5,7 +5,7 @@ import { config } from './config'
 export function postTestClientStart (): Response {
   return http.post(
     buildUrl('start', config.testClientExecuteUrl),
-    JSON.stringify({ frontendUri: config.frontendUrl }),
+    JSON.stringify({ backendUri: `${config.backendUrl}/`, frontendUri: config.frontendUrl }),
```
jamielutton-gds, line 8: `backendUri` is not the name of the key that needs to be passed in the test client body to change the ApiLocation. The key name should be `target`

jamielutton-gds: Passing in `backendUri` here will not override the value in the ApiLocation, which means that the ApiLocation in the response will still use the default non-pretty BE URL. There is an issue with passing this into the test client at the moment. 

When we pass in a `target` which does not contain a `/` at the end, we get a response from the test client, but the ApiLocation is missing a forward slash before verifyAuthorizeRequest, e.g. `https://api-backend-api.review-b.build.account.gov.ukverifyAuthorizeRequest`.

When the `target` does not contain a `/` at the end, there is an error in invoking one of the APIs that the test client lambda calls (I think it's the /.well-known/jwks.json endpoint). 

So, if we include a / we know we'll get an error from the test client. Which means if we want to complete this work without updating the test client we would need to pass in `target: config.backendUrl` to the body here, and tactically edit the ApiLocation so that we insert a / between the `backendUrl` and the `verifyAuthorizeRequest`, or something along these lines. 

Open to other suggestions but this is my understanding of the situation at the moment!

deploy/scripts/src/mobile/utils/test-client.ts
```diff
@@ -19,5 +19,5 @@ export function parseTestClientResponse (response: Response, location: 'WebLocat
     throw new Error('Failed to parse URL from response')
   }
 
-  return url
+  return url.replace(config.backendUrl, `${config.backendUrl}/`)
```
jamielutton-gds, line 22: Another potential option which might be preferred over using replace would be to look at what we do in the TESTCLIENT_START test script in the Postman collection? The rationale for this approach would be that if we were to fix the test client bug, we wouldn't be required to make changes to the performance test repo. 

In the Postman test script, we parse the ApiLocation to extract the query parameters, and then append these to `<backendUrl>/verifyAuthorizeRequest?`. Are you able to have a look into doing it this way? Happy to talk through with you if any of that is confusing!

raluca-alecu: In the Postman collection, the request_jwt and client_id were added as environment variables, not from the test client response. 
I think we should update the bug so we'll know that we should remove the replace from this method when fixed instead of adding a different implementation as a workaround for a bug 🤔

jamielutton-gds: The request_jwt and client_id are set here, after we make a request to the test client: 
image

We do have default values in the env vars, but we override them with the test client response's ApiLocation

deploy/scripts/src/mobile/utils/test-client.ts
```diff
@@ -1,23 +1,54 @@
...
+export function parseApiLocation (apiLocation: string): string {
+  const queryParams = apiLocation.split('?')[1].split('&')
+  const requestJwt = queryParams
+    .filter((value) => value.startsWith('request'))[0]
+    .split('=')[1]
+  const clientId = queryParams
+    .filter((value) => value.startsWith('client_id'))[0]
+    .split('=')[1]
+
+  const apiLocationUrl = buildBackendUrl('verifyAuthorizeRequest', {
```
jamielutton-gds, line 47: Here we can just return the output of `buildBackendUrl` inline, i.e. `return buildBackendUrl(...)`

deploy/scripts/src/mobile/utils/test-client.ts
```diff
@@ -1,23 +1,54 @@
...
+  }
+
   if (typeof url !== 'string') {
     throw new Error('Failed to parse URL from response')
   }
 
   return url
 }
+
+export function parseApiLocation (apiLocation: string): string {
```
jamielutton-gds, line 38: I don't think this function needs to be exported

deploy/scripts/src/mobile/utils/test-client.ts
```diff
@@ -1,23 +1,54 @@
...
-export function parseTestClientResponse (response: Response, location: 'WebLocation' | 'ApiLocation'): string {
-  const url = response.json(location)
+export function parseTestClientResponse (
+  response: Response,
+  location: 'WebLocation' | 'ApiLocation'
+): string {
+  let url
+  if (location === 'ApiLocation') {
+    const apiLocation = response.json(location) as string
+    url = parseApiLocation(apiLocation)
```
jamielutton-gds, line 26: By putting `parseApiLocation` before the `if (typeof url !== 'string') {` check, we're changing the behaviour of the error handling of this function. I think the `if (location === 'ApiLocation')` check should come after the `if (typeof url !== 'string')`, i.e. something like this:
```
const url = response.json(location)
if (typeof url !== 'string') {
    throw new Error('Failed to parse URL from response')
}
if (location === 'ApiLocation') {
  return parseApiLocation(url)
} else {
  return url
}
```

jamielutton-gds (approved):

deploy/scripts/src/mobile/utils/test-client.ts
```diff
@@ -1,23 +1,48 @@
...
+    return parseApiLocation(url)
+  } else {
+    return url
+  }
+}
+
+function parseApiLocation (apiLocation: string): string {
+  const queryParams = apiLocation.split('?')[1].split('&')
+  const requestJwt = queryParams
+    .filter((value) => value.startsWith('request'))[0]
```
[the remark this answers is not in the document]
marija8327: Updated

deploy/scripts/src/mobile/utils/test-client.ts
marija8327: Updated

PR #179, opened 2023-07-04
[QA-247]: Refactored Orange CRI scripts

abhinaysrivastavaperf:
### What?
Refactored Orange CRI scripts to reduce the size of the single script.

#### Changes:
- Split KBV and Address scripts in different folders
- Added 'utils' folder for functions for 'assertion' and 'environment variables'
- Used functions calls for assertions in both the scripts.

---

### Why?

- The objective was to reduce the size of single script and split it as per CRI (address and kbv) as requested by Orange CRI team
- Also, this approach will be beneficial considering there will be more CRIs in the future.

---

### Related:
- [OJ-1634](https://govukverify.atlassian.net/browse/OJ-1634)

[OJ-1634]: https://govukverify.atlassian.net/browse/OJ-1634?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


deploy/scripts/src/cri-orange/address/test.ts
```diff
@@ -262,14 +84,9 @@ export function addressScenario1 (): void {
         }
       )
       const endTime = Date.now()
-
-      check(res, {
-        'is status 200': (r) => r.status === 200,
-        'verify page content': (r) => (r.body as string).includes('Find your address')
-      })
+      isStatusCode200(res) && validatePageContent(res, 'Find your address')
```
Tom-Dann, line 87: The short-circuit evaluation here will not call the second check if the status code is not 200. I think the desired behaviour should be to evaluate both checks before exiting the iteration on an assertion failure.

abhinaysrivastavaperf: If the status code is not 200 then validating the page content would not be required I think.

Tom-Dann: Probably one to just be conscious of in the metrics, this will mean that the page content validation only ever runs after a successful assertion on the status code. Resulting in less failing page validation checks when we get errors.

surajkumar: It would be a good idea to validate the page contents. If the behaviour changes and the page returns something unexpected, the test should at least fail.

abhinaysrivastavaperf: Yes the page contents are being validated. Basically this statement is first validating whether the final HTTP status code of the response is 200 OK or not. Then it is validating whether the page content of the response is as expected or not.

merlinc (approved):
LGTM

PR #184, opened 2023-07-06
Qa 253

paulfloodgds:
### What?
Updated TXMA and BTM Smoke and Load Test Profiles

#### Changes:
- Increased duration of smoke tests to 60 secs and added Load Profiles to cover initial performance baselines.

---

### Why?
Smoke test running successfully locally but failing in pipeline due to timeout, suggested change is to increase the duration.

### DIFF
--- a/deploy/scripts/src/txma/test.ts
+++ b/deploy/scripts/src/txma/test.ts
@@ -11,7 +11,7 @@ const profiles: ProfileList = {
       preAllocatedVUs: 1,
       maxVUs: 2,
       stages: [
-        { target: 1, duration: '5s' } // Ramps up to target load
+        { target: 1, duration: '60s' } // Ramps up to target load
       ],
       exec: 'sendEvent'
     }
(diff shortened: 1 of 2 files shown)


deploy/scripts/src/txma/test.ts
```diff
@@ -11,7 +12,64 @@ const profiles: ProfileList = {
...
+      exec: 'sendEvent'
+    },
+    sendEventLoadTest4: {
+      executor: 'ramping-arrival-rate',
+      startRate: 1,
+      timeUnit: '1s',
+      preAllocatedVUs: 1,
+      maxVUs: 1500,
+      stages: [
+        { target: 30, duration: '60m' } // Ramp up to 30 iterations per second in 60 minutes
```
Tom-Dann, line 72: This scenario is the same as `sendEventLoadTest3` above, is this supposed to be a higher value?

Tom-Dann: Never mind - just noticed to difference in duration

Tom-Dann (approved):

deploy/scripts/src/txma/test.ts
paulfloodgds: Yes 10min and 60mins

andylangridge (approved):

markschnitziusgovuk (approved):

PR #209, opened 2023-07-18
[QA-269] Journey was split in two scenarios

gpopescu23:
### What?
Journey was split in two scenarios

#### Changes:
- We have the first journey for IPVCore and the second for Alpha
- IPVCore is for persisting the data and Alpha si for retriving

---

### Why?
They are two different journeys and need to be tested separately


deploy/scripts/src/accounts/id-reuse.ts
```diff
@@ -76,7 +116,7 @@ const env = {
   envApiKey: __ENV.ACCOUNT_BRAVO_ID_REUSE_API_KEY,
   envApiKeySummarise: __ENV.ACCOUNT_BRAVO_ID_REUSE_API_KEY_SUMMARISE
 }
-export function idReuse (): void {
+export function idReuseIPVCore (): void {
```
Tom-Dann, line 119: These should be renamed more appropriately - both scenarios originate from IPV Core.

Consider splitting them by 'persist ID' and 'retrieve ID'

Tom-Dann (approved):

PR #217, opened 2023-07-20
DCMAW-6068: Mobile - add custom duration Trend for FE E2E script

raluca-alecu:
### What?
Added transactionDuration equal with request endTime - startTime for FE E2E test

#### Changes:
- all methods used in the FE E2E test: capturing startTime/endTime before/after request and if assertions are passing changed transactionDuration to equal request endTime - startTime for FE E2E test

---

### Why?
We are interested in the total duration of all requests, including redirects, so in our scripts, we have been adding a custom response time duration metric to track this.

Duration:
Screenshot 2023-07-20 at 16 57 32

### DIFF
--- a/deploy/scripts/src/mobile/testSteps/backend.ts
+++ b/deploy/scripts/src/mobile/testSteps/backend.ts
@@ -66,7 +66,7 @@ export function getBiometricTokenV2 (sessionId: string): void {
 
     isStatusCode200(res)
       ? transactionDuration.add(endTime - startTime)
-      : fail('Response Validation Failed')
+      : fail()
   })
 }
 
@@ -88,7 +88,7 @@ export function postFinishBiometricSession (sessionId: string): void {
 
     isStatusCode200(res)
       ? transactionDuration.add(endTime - startTime)
-      : fail('Response Validation Failed')
+      : fail()
   })
 }
 
@@ -107,7 +107,7 @@ export function getRedirect (sessionId: string): {
 
     isStatusCode200(redirectRes)
       ? transactionDuration.add(endTime - startTime)
-      : fail('Response Validation Failed')
+      : fail()
 
     return {
       authorizationCode: redirectRes.json('authorizationCode') as string,
--- a/deploy/scripts/src/mobile/testSteps/backend.ts
+++ b/deploy/scripts/src/mobile/testSteps/backend.ts
@@ -1,5 +1,5 @@
 import http from 'k6/http'
-import { group } from 'k6'
+import { fail, group } from 'k6'
 import { uuidv4 } from '../../common/utils/jslib/index'
 import { buildBackendUrl } from '../utils/url'
 import {
@@ -7,6 +7,9 @@ import {
   postTestClientStart
 } from '../utils/test-client'
 import { isStatusCode200, isStatusCode201 } from '../utils/assertions'
+import { Trend } from 'k6/metrics'
+
+const transactionDuration = new Trend('duration')
 
 export function postVerifyAuthorizeRequest (): string {
   let verifyUrl: string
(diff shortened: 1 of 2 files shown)


Tom-Dann (approved):
The code looks good to me, but I think the screenshot in the evidence should have a `duration` row if it includes these changes

deploy/scripts/src/mobile/testSteps/backend.ts
```diff
@@ -54,10 +57,16 @@ export function getBiometricTokenV2 (sessionId: string): void {
     const biometricTokenUrl = buildBackendUrl('/biometricToken/v2', {
       authSessionId: sessionId
     })
+
+    const startTime = Date.now()
```
sandy-may, line 61: Why do we need it for this request? This is a BE API that returns a 2XX for a happy scenario (no redirect).

@Tom-Dann - do we need to be consistent with this timing approach, i.e. apply it to every single request, or can we only apply it where we know we have redirects, and want to track the overall request time.

deploy/scripts/src/mobile/testSteps/backend.ts
```diff
@@ -54,10 +57,16 @@ export function getBiometricTokenV2 (sessionId: string): void {
...
     const res = http.get(biometricTokenUrl, {
       tags: { name: 'GET /biometricToken/v2' }
     })
+    const endTime = Date.now()
+
     isStatusCode200(res)
+      ? transactionDuration.add(endTime - startTime)
+      : fail('Response Validation Failed')
```
sandy-may, line 69: @raluca-alecu - what does the fail() do here? Does it stop our test from proceeding to the next step if a request fails?

deploy/scripts/src/mobile/testSteps/backend.ts
Tom-Dann: It makes sense to do it for every request, then you have one metric that can capture the total duration regardless of if it has redirects or not

deploy/scripts/src/mobile/testSteps/backend.ts
raluca-alecu: Immediately throw an error, aborting the current script iteration.

sandy-may: That's great (and we probably wanted it there in place before, otherwise we'd get a whole bunch of failures if an early step failed) thank you.

sandy-may: Do we need an argument into the fail function? I can't work out it's value here, but might be missing something. I.e. can we simply go fail()

raluca-alecu: 1. The argument is optional. I followed the implementation of fail in other tests from this repo.
2. If we calculate the error request duration, this will be displayed on the final results, and our max request time and average may be affected 🤔 
Maybe @Tom-Dann can give his opinion 🙏

deploy/scripts/src/mobile/testSteps/backend.ts
Tom-Dann: They are all sent by default, so this one in particular will be accessible as `k6.duration` 
image

deploy/scripts/src/mobile/testSteps/backend.ts
Tom-Dann: It's not very helpful to include failing requests in the response time metrics, so this implementation is good (and also aligns to what we are doing in other test scripts). Our response time NFRs should be based on passing requests only.

An alternative in the future may be to track the duration of every request and then add a tag based on whether the assertions passed. Similar to how the default `http_req_duration` has a `expected_response` boolean tag on every metric. That would still give you the option to filter on successful requests, and also potentially some insight into durations for errors too (e.g. if it errors and the duration is 30,000ms it's easy to see it was probably a timeout)

For now though, tracking this custom metric on only passing requests suits our needs IMO

sandy-may: Yep I agree on using success requests for metrics.

PR #218, opened 2023-07-20
[QA-277] Retrieve static resources in IPV Core Script

Tom-Dann:
### What?
Adding a utility to programmatically retrieve static resources from a page, and implemented in the IPV Core script

#### Changes:
- Added a utility file `common/utils/request/static.ts`. This contains a function which, when given a `Response`, it will pull all `src` and `href` values and call retrieve them using `http.batch()`. This simulates the behaviour of a browser resolving and retrieving static resources such as `.css` stylesheets and `.js` scripts.
- Added a `request/static` group to the `common/unit-tests.ts` to test that the `resolveURL()` function performs as expected
- Added calls in the IPV Core script to retrieve static resources after every expected 200 response page

---

### Why?
Currently we are only testing the delivery of dynamic resources, and would like to add static resources to scripts to test how changes to CDNs and static asset delivery performs

---

### Related:
- [`http.batch()` documentation](https://k6.io/docs/javascript-api/k6-http/batch/)


deploy/scripts/src/ipv-core/test.ts
```diff
@@ -119,6 +121,7 @@ export function coreScenario1 (): void {
           tags: { name: 'B01_Core_01_LaunchOrchestratorStub' }
         }
       )
+      if (env.staticResources) getStaticResources(res)
```
abhinaysrivastavaperf, line 124: Is it possible to include some kind of `check` to validate that the `http.batch` request was successful?

abhinaysrivastavaperf: As discussed, adding a check specifically for the http.batch request would be too noisy. If any of the request fail, it would be added to the http_req_failed metric.
Closing the comment.

abhinaysrivastavaperf (approved):

TobySaundersGDS:
Hi Tom - only just go around to looking at this, looks like it will download anything directlyy referenced in the html, but if there are other files referenced in the CSS it will not try to acquire them - is there a way to make it recursive if the file it pulls has a .css extension?

PR #180, opened 2023-07-04
[QA-245] Add SPoT permissions to CodeBuild policy

Tom-Dann:
### What?
Adding permissions to the CodeBuild policy to allow the agent to call SQS queues and lambdas in the SPoT Build environment

#### Changes:
- Added a mapping for the SPoT Build environment AWS ID
- Added policy to access KMS keys relating to Request and Output SQS Queues in the SPoT Build environment
- Added policy to interact with Request and Output SQS Queues in the SPoT Build environment
- Added policy to allow invoking of lambdas in the SPoT Build environment
- Picked up some file formatting changes

### Why?
To support testing of SPoT scenarios, and directly creating/reading SQS events and lambda 

---

### Related:
- #174 
- [**SPT-273**](https://govukverify.atlassian.net/browse/SPT-273)
- [**SPT-277**](https://govukverify.atlassian.net/browse/SPT-277)

[SPT-273]: https://govukverify.atlassian.net/browse/SPT-273?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SPT-277]: https://govukverify.atlassian.net/browse/SPT-277?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #220, opened 2023-07-21
[QA-261] Add execution role to enable SQS testing

Tom-Dann:
### What?
Add a new execution role which will be used in scripts as credentials to send SQS events to test SQS queues and services which use SQS as an entrypoint

#### Changes:
- `deploy/template.yaml`:
  - Added a new execution role `${AWS::StackName}-PerformanceTesterRole` which will be the role used within the test scripts to call SQS queues. The permissions have been split out so now all the relevant execution permissions are on this role (SQS, Lambda, KMS)
  - Renamed the CodeBuild Service role from `${AWS::StackName}-PerformanceTester` to `${AWS::StackName}-PerformanceServiceRole`
  - Added a step to the pre-build to export the assume-role to an environment variable, for use in scripts
- `.../spot/test.ts` & `.../txma/test.ts`: Updated the script to use the new execution credentials
- `deploy/scripts/src/common/utils/aws/types.d.ts`: Added a new `interface` type to handle parsing of the expected AWS CLI JSON format

---

### Why?
This is to enable SQS testing from our performance testing pipeline, and programmatically pulling credentials into the script to enable testing

---

### Related:
- #180

### DIFF
--- a/deploy/scripts/src/common/utils/aws/types.d.ts
+++ b/deploy/scripts/src/common/utils/aws/types.d.ts
@@ -1,0 +1,13 @@
+// AWS CLI Credential Output Format https://docs.aws.amazon.com/cli/latest/reference/sts/assume-role.html#examples
+export interface AssumeRoleOutput {
+  'AssumedRoleUser': {
+    'Arn': string
+    'AssumedRoleId': string
+  }
+  'Credentials': {
+    'AccessKeyId': string
+    'Expiration': string
+    'SecretAccessKey': string
+    'SessionToken': string
+  }
+}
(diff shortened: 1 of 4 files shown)


deploy/scripts/src/common/utils/aws/types.d.ts
```diff
@@ -0,0 +1,13 @@
+// AWS CLI Credential Output Format https://docs.aws.amazon.com/cli/latest/reference/sts/assume-role.html#examples
+export interface AssumeRoleOutput {
```
chris-dco, line 2: Rather than defining the type I'd recommend importing the AWS Client STS library `@aws-sdk/client-sts` and use `AssumeRoleCommandOutput`. Also you don't need to have quotes `'...'` for definitions.

chris-dco (changes requested):

deploy/scripts/src/common/utils/aws/types.d.ts
Tom-Dann: Thanks, I've removed the single quotes in the interface definition.

The k6 scripts are not strictly TS/JS where you can important any library. Instead we can only import `k6` libraries and small files written in TS/JS. So having a small type definition file for this interface is the simplest solution here.

chris-dco (approved):

fpmrqs (approved):

abhinaysrivastavaperf (approved):

PR #245, opened 2023-08-01
DCMAW-5597: Added load profile 

raluca-alecu:
### What?
Added a load profile to the authorize test accordingly with the ones in the ticket:
- A 15 minutes volume ramp up from 1 request per second to 100 requests per second (0.1 t/s/s NFR)
- A 30 minutes period of sustained volume at 100 requests per second

#### Changes:
- authorize.ts: added load profile

---

### Why?
The full FE E2E test will test the performance of the FE endpoints, simulating concurrent users following the Frontend journey.

---

### DIFF
--- a/deploy/scripts/src/mobile/authorize.test.ts
+++ b/deploy/scripts/src/mobile/authorize.test.ts
@@ -26,10 +26,10 @@ const profiles: ProfileList = {
   load: {
     startJourney: {
       executor: 'ramping-arrival-rate',
-      startRate: 0,
+      startRate: 1,
       timeUnit: '1s',
       preAllocatedVUs: 50, // Calculation: 100 journeys / second * 0.5 seconds average journey time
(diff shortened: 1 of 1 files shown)


deploy/scripts/src/mobile/authorize.test.ts
```diff
@@ -22,6 +22,20 @@ const profiles: ProfileList = {
...
       exec: 'authorize'
     }
+  },
+  load: {
+    startJourney: {
+      executor: 'ramping-arrival-rate',
+      startRate: 0,
+      timeUnit: '1s',
+      preAllocatedVUs: 50, // Calculation: 100 journeys / second * 0.5 seconds average journey time
+      maxVUs: 200, // Calculation: 100 journeys / second * 2 seconds maximum journey time
```
Tom-Dann, line 32: I'd recommend bumping this up to 300, so it can support just over the max expected 2.5 second response time (from NFRs). Hopefully won't be needed though

sandy-may (approved):

PR #248, opened 2023-08-02
[QA-287]: Fix Audience ID and Client ID

abhinaysrivastavaperf:
### What?
Fix audience ID and client ID values in SPOT request

#### Changes:
- According to logic in SPOT validation function, the value of `out_audience` should be equal to the value of `log_ids.client_id`. 

---

### Why?
To fix the errors in SPOT request processing in build environment.

### DIFF
--- a/deploy/scripts/src/spot/requestGenerator/spotReqGen.ts
+++ b/deploy/scripts/src/spot/requestGenerator/spotReqGen.ts
@@ -1,9 +1,8 @@
 import { type SpotRequest } from '../../spot/requestGenerator/spotReqFormat'
 import { uuidv4 } from '../../common/utils/jslib/index'
 
-const audClientID = uuidv4()
-
 export function generateRequest (currTime: string): SpotRequest {
+  const audClientID = uuidv4()
   const sampleSpotRequest: SpotRequest = {
     in_claims: {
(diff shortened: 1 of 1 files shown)


deploy/scripts/src/spot/requestGenerator/spotReqGen.ts
```diff
@@ -1,6 +1,8 @@
 import { type SpotRequest } from '../../spot/requestGenerator/spotReqFormat'
 import { uuidv4 } from '../../common/utils/jslib/index'
 
+const audClientID = uuidv4()
+
 export function generateRequest (currTime: string): SpotRequest {
```
Tom-Dann, line 6:
```suggestion
export function generateRequest (currTime: string): SpotRequest {
  const audClientID = uuidv4()
```
Move the variable declaration inside the function so there is a new UUID for every iteration

Tom-Dann (changes requested):

deploy/scripts/src/spot/requestGenerator/spotReqGen.ts
abhinaysrivastavaperf: Done

Tom-Dann (approved):

PR #266, opened 2023-08-11
[QA-298] Use `submitForm()` in CRI Kiwi script

Tom-Dann:
### What?
Using `Response.submitForm()` in place of the `http.post()` where applicable, and updated test scenarios with new pages.

This has been validated with smoke tests locally

#### Changes:
- Replaced `http.post()` instances with `Response.submitForm()` in places where we are submitting the form from the previous request response. Also removed any input values which are set to the default.
- Removed unused `getCSRF()` and `getRequestCode()` utility functions
- Removed some unneeded sleep times
- Added an extra step to FaceToFace NationalIDEEA, EU-DL and Non-UKPassport branches to confirm expiry date exists
- Split out the CIC and F2F environment variables, as they use different values for the stub URL and target. This has been updated in the `template.yaml` too

---

### Why?
To simplify the test script and reduce unnecessary hard-coded form input values

---

### Related:
- [`Response.submitForm()` documentation](https://k6.io/docs/javascript-api/k6-http/response/response-submitform/)


Tom-Dann:
The CIC journey is currently broken and needs to be updated before merging this PR

abhinaysrivastavaperf (dismissed):

abhinaysrivastavaperf:
> The CIC journey is currently broken and needs to be updated before merging this PR

Missed this comment while reviewing the PR. This will not be merged until the CIC journey is validated

abhinaysrivastavaperf (approved):

PR #314, opened 2023-09-13
[QA-223] : Added a separate approach for accessing parameter store

abhinaysrivastavaperf:
### What?
Alternative approach for accessing the SSM parameter store to fetch the value of the variables

#### Changes:
- Added the `aws-ssm` utility in `./src/common/utils/jslib` folder and the corresponding declaration file.
- Created a separate test file `ssmTest` to test the approach separately in dev environment. This isolated file will be removed and the approach will be implemented in individual scripts before the PR is ready for review.
- Added the SSM related permissions to the `PerformanceTesterPolicy` to be used by the `PerformanceTesterRole` in CF template.
- Removed the entry from BuildSpec in the CF template.

---

### Why?
Alternative approach is required to reduce changes in the CF template in case more environment variables are added.


abhinaysrivastavaperf:
Closing this PR as the proposed solutions are more complex and require more maintenance efforts than the existing solution.

PR #329, opened 2023-09-25
[QA-337] : Updated test data approach in IPV core

abhinaysrivastavaperf:
### What?
Updated test data approach in IPV core to optimise load injector memory utilisation

#### Changes:
- Removed test data approach of using JSON files.
- Created variables for tests data payloads as this is constant for each virtual user.

---

### Why?
Old test data approach was using open function to read the JSON file. Open function is memory intensive as each virtual user keeps a separate copy of the file in memory. This is causing high memory utilisation of the load injector resulting in failed tests.


Tom-Dann (approved):
Might want to include types from [`alphagov/di-identity-vocab`](https://github.com/alphagov/di-identity-vocab) in the future for some of this data

PR #348, opened 2023-10-10
Updated IPVR script with updated payloads

sachadorf13:
### What?
Updated IPVR script according to updated payloads

#### Changes:
- Added `F2F_DOCUMENT_UPLOADED` event payload
- Updated `F2F_YOTI_START` and `IPV_F2F_CRI_VC_CONSUMED` payloads with new fields

---

### Why?
Team Kiwi has made changes to these payloads so our scripts need to be updated accordingly

---

### DIFF
--- a/deploy/scripts/src/cri-kiwi/requestGenerator/ipvrReqFormat.ts
+++ b/deploy/scripts/src/cri-kiwi/requestGenerator/ipvrReqFormat.ts
@@ -32,8 +32,8 @@ export interface F2F_YOTI_START {
         address: string
         location: [
           {
-            latitude: string
-            longitude: string
+            latitude: number
+            longitude: number
           }
         ]
         postcode: string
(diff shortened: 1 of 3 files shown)


deploy/scripts/src/cri-kiwi/requestGenerator/ipvrReqFormat.ts
```diff
@@ -25,6 +25,44 @@ export interface F2F_YOTI_START {
...
     govuk_signin_journey_id: string
   }
+  extensions: {
+    post_office_details: [
+      {
+        name: string
+        address: string
+        location: [
+          {
+            latitude: string
```
abhinaysrivastavaperf, line 35: This should be a `number`

abhinaysrivastavaperf (changes requested):

abhinaysrivastavaperf (approved):

PR #361, opened 2023-10-17
[QA-268]: Added custom metric to track iterations

abhinaysrivastavaperf:
### What?
Custom metric to track number of iterations in a test

#### Changes:
- Added custom metrics Iterations_Started and Iterations_Completed in the IPV core script.
- Added a value to the Iterations_Started metric before the start of the first transaction.
- Added a value to the Iterations_Completed metric after the check of the last transaction.

---

### Why?
To track the number of iterations started and completed in a test.

### DIFF
--- a/deploy/scripts/src/common/utils/custom_metric/counter.ts
+++ b/deploy/scripts/src/common/utils/custom_metric/counter.ts
@@ -1,0 +1,4 @@
+import { Counter } from 'k6/metrics'
+
+export const iterationsStarted = new Counter('Iterations_Started')
+export const iterationsCompleted = new Counter('Iterations_Completed')
--- a/deploy/scripts/src/common/utils/custom_metric/counter.ts
+++ b/deploy/scripts/src/common/utils/custom_metric/counter.ts
@@ -1,4 +1,4 @@
 import { Counter } from 'k6/metrics'
 
-export const iterationsStarted = new Counter('Iterations_Started')
-export const iterationsCompleted = new Counter('Iterations_Completed')
+export const iterationsStarted = new Counter('iterations_started')
+export const iterationsCompleted = new Counter('iterations_completed')
--- a/deploy/scripts/src/ipv-core/test.ts
+++ b/deploy/scripts/src/ipv-core/test.ts
@@ -1,3 +1,4 @@
+import { Counter } from 'k6/metrics'
 import encoding from 'k6/encoding'
 import { group } from 'k6'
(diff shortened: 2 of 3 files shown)


deploy/scripts/src/ipv-core/test.ts
```diff
@@ -94,6 +95,8 @@ const profiles: ProfileList = {
 }
 
 const loadProfile = selectProfile(profiles)
+const iterationsStarted = new Counter('Iterations_Started')
+const iterationsCompleted = new Counter('Iterations_Completed')
```
Tom-Dann, line 99: I think it would make sense to move these to a common utilities file and then import them. That way the metric name will be consistent across all tests that use these

Tom-Dann (changes requested):

deploy/scripts/src/ipv-core/test.ts
abhinaysrivastavaperf: Done

deploy/scripts/src/common/utils/custom_metric/counter.ts
```diff
@@ -0,0 +1,4 @@
+import { Counter } from 'k6/metrics'
+
+export const iterationsStarted = new Counter('Iterations_Started')
```
Tom-Dann, line 3: Add a check in `unit-tests.ts` for these utilities, to check they are behaving as expected

abhinaysrivastavaperf: Added a test for custom metric check. We cannot get the value of a counter from inside a k6 test so I have added a threshold check on the counters.

deploy/scripts/src/common/utils/custom_metric/counter.ts
```diff
@@ -0,0 +1,4 @@
...
+export const iterationsCompleted = new Counter('Iterations_Completed')
```
Tom-Dann, line 4:
```suggestion
export const iterationsStarted = new Counter('iterations_started')
export const iterationsCompleted = new Counter('iterations_completed')
```
Suggest making them lowercase in line with other k6 metrics, but otherwise okay

Tom-Dann (approved):

deploy/scripts/src/common/utils/custom_metric/counter.ts
abhinaysrivastavaperf: Done, thank you

PR #411, opened 2023-12-07
[DCMAW-7214] Add development environment config for Mobile tests

Tom-Dann:
### What?
Adding config to enable switching between Build and Development environments for Mobile tests

#### Changes:
- `deploy/template.yml`:
  - Split mobile environment variables into BUILD and DEV, one copy for each environment
  - Sorted all environment variables alphabetically - this affected some unrelated variables
- `deploy/scripts/src/mobile/utils/config.ts`: Added logic to select environment variables based on the environment, and added a check to make sure a valid environment is provided

---

### Why?
To enable testing against the development environment for mobile tests

PR #425, opened 2023-12-12
[QA-389] Add development environment to IPV Core script

Tom-Dann:
### What?
Adding config to enable switching between Build and Development environments for the IPV Core test

#### Changes:
- `deploy/template.yml`: Split IPV Core URL environment variable into BUILD and DEV
- `deploy/scripts/src/mobile/utils/config.ts`: Added logic to select environment variables based on the environment, and added a check to make sure a valid environment is provided

---

### Why?
To facilitate testing of canary deployments, we need to send traffic to the development environment using the IPV Core script

---

### Related:
- #411


beca-galliano (approved):

abhinaysrivastavaperf (approved):
Below parameters will be need to be added in the parameter store (if not already done)

```
/perfTest/identity/build/coreUrl
/perfTest/identity/dev/coreUrl
```

Below parameter can be deleted
`/perfTest/identity/coreUrl`

PR #453, opened 2024-01-11
DCMAW-7817: Added deploy scenario for mobile FE tests

raluca-alecu:
### What?
The smoke scenario details for the FE E2E mobile scripts. 

#### Changes:
- target of 1 request per second 
- 25m sustained volumes at 1 requests per seconds
- preAllocated VUs decreased to 15 
---

### Why?
Migrate a Frontend stack in the Dev account from the DCMAW to Dev Platform VPC whilst a performance test is running. It is suggested that this performance test runs at low volumes, aiming at 1TPS per journey

---

### Related:
- [DCMAW-7817](https://govukverify.atlassian.net/browse/DCMAW-7817) 

EVIDENCE:
Screenshot 2024-01-11 at 17 19 00

[DCMAW-7817]: https://govukverify.atlassian.net/browse/DCMAW-7817?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/deploy/scripts/src/mobile/backend-e2e.test.ts
+++ b/deploy/scripts/src/mobile/backend-e2e.test.ts
@@ -39,19 +39,6 @@ const profiles: ProfileList = {
       ],
       exec: 'backendJourney'
     }
-  },
-  deploy: {
-    backendJourney: {
-      executor: 'ramping-arrival-rate',
-      startRate: 1, // start with one iteration
-      timeUnit: '1s',
-      preAllocatedVUs: 10, // Calculation: 1 journeys / second * 10 seconds average journey time
-      maxVUs: 65, // Calculation: 1 journeys / second * 14 seconds maximum journey time + 50 buffer
-      stages: [
-        { target: 1, duration: '25m' } // maintain 1 iterations per second for 25 min
-      ],
-      exec: 'backendJourney'
-    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


Tom-Dann:
I'd suggest maybe keeping the `smoke` load profile, and adding a new one for this use case called `deployment` or similar - if it is to be used in deployment testing

jamesamooney:
> I'd suggest maybe keeping the `smoke` load profile, and adding a new one for this use case called `deployment` or similar - if it is to be used in deployment testing

Was going to ask/suggest similar. With the current implentation would the plan be to raise another PR once the work is done for [DCMAW-7817](https://github.com/govuk-one-login/performance-testing/pull/453#top) to revert this change? 

If we did what @Tom-Dann mentioned and made a new profile, would we then just leave this profile in the codebase or would we still need to remove it?

raluca-alecu:
I'm happy with adding a new profile. In this case, @Tom-Dann, what else do we need to change to be able to run the scenario from the aws code build? 
@jamesamooney I assume we can leave the newly created scenario.

Tom-Dann:
> @Tom-Dann, what else do we need to change to be able to run the scenario from the aws code build?
It'll just be the same as any other test where you specify the load profile as the value for the `PROFILE` environment variable. So instead of the default value `smoke` it'll be whatever you call this new one

sandy-may (approved):
Assuming this profile works when running locally, happy with this

jamesamooney (approved):

raluca-alecu:
WIP testing with this profile. I have some issues not related to the new profile added. I'll post here the evidence when done.

PR #476, opened 2024-01-30
[QA-408]: Added scripts for AIS scenarios

abhinaysrivastavaperf:
### What?
Added scripts for AIS scenarios

#### Changes:
- Added permissions for `PerformanceTester` role in CF template.
- Added `Persist Intervention` scenario.
- Added `Retrieve Intervention` scenario

---

### Why?
To conduct AIS performance testing


deploy/template.yaml
```diff
@@ -40,9 +40,12 @@ Mappings:
   ContraIndicators:
     AWS:
       AccountID: "755415363251"
-  TxMA:
+  TxMABuild:
     AWS:
       AccountID: "750703655225"
+  TxMAStaging:
+    AWS:
+      AccountID: "178023842775"
```
Tom-Dann, line 48:
```suggestion
  TxMA:
    Build:
      AccountID: "750703655225"
    Staging:
      AccountID: "178023842775"
```

deploy/scripts/src/ais/requestGenerator/aisReqFormat.ts
```diff
@@ -0,0 +1,19 @@
+export interface TICF_ACCOUNT_INTERVENTION {
```
Tom-Dann, line 1:
```suggestion
export interface TicfAccountIntervention {
```
Prefer PascalCase for types/interfaces

deploy/scripts/src/ais/test.ts
```diff
@@ -0,0 +1,158 @@
...
+      exec: 'persistIV'
+    },
+    retrieveIV: {
+      executor: 'ramping-arrival-rate',
+      startRate: 1,
+      timeUnit: '1s',
+      preAllocatedVUs: 1,
+      maxVUs: 11400,
+      stages: [
+        { target: 3800, duration: '15m' }, // Ramp up to 3800 iterations per second in 15 minutes
```
Tom-Dann, line 88: Should this be 1900 and two calls in the scenario?

deploy/template.yaml
abhinaysrivastavaperf: Done

deploy/scripts/src/ais/requestGenerator/aisReqFormat.ts
abhinaysrivastavaperf: Done

deploy/scripts/src/ais/test.ts
abhinaysrivastavaperf: No the worst case scenario would be the peak volume of calls from Auth (1900/s) and Core (1900/s) simultaneously. The GET call from both Auth and Core is same so testing this at 3800/s would be the worst case peak volume of this GET call.

deploy/scripts/src/ais/requestGenerator/aisReqFormat.ts
Tom-Dann, line 1: Makes sense to move this interface definition to the same file as `deploy/scripts/src/ais/requestGenerator/aisReqGen.ts`

Tom-Dann (approved):

PR #480, opened 2024-02-01
[QA-405] Start auth journey without stub RP

mrwilson:
### What?
Start Authentication journeys at the GOV.UK One Login provider directly rather than going through the stub RP

New variables required:

- `ACCOUNT_OP_URL`: The /authorize endpoint of the One Login environment e.g.`https://oidc.<environment>.account.gov.uk/authorize`
- `ACCOUNT_RP_STUB_CLIENT_ID`: The client ID of the stub RP which provides the callback URL at the end of the journey

#### Changes:
- Create a URL for directly starting a One Login journey
- Dispatch signIn and signUp journeys using new URL rather than stub
- Add new parameter store variables to BuildSpec
---

### Why?
The stub RP is not productionised to the extent that it can handle being the entry point for a large number of Authentication sign-in/sign-up journeys. This PR creates a GOV.UK One Login /authorize request directly and bypasses the stub when starting a journey.

## Pre-merge check

- [ ] Paramater Store values created in `performance-test-prod` account


Tom-Dann (approved):
Ensure the two new parameters are created in the parameter store before merging

PR #485, opened 2024-02-05
[QA-419]: Update S3 property to fix issue

abhinaysrivastavaperf:
### What?
Update S3 results bucket property to fix deployment issue.

#### Changes:
- Added `OwnershipControls` property to replace the `AccessControl` property.
- This has been tested in the dev environment.

---

### Why?
To fix the deployment issue in perf test accounts.

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -324,6 +324,9 @@ Resources:
   ResultsBucket:
     Type: AWS::S3::Bucket
     Properties:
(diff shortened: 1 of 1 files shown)


Tom-Dann (changes requested):
This changes the ownership policy and also fails to deploy
`The bucket does not allow ACLs`

Tom-Dann:
Replaced by #486

PR #517, opened 2024-03-01
Added ts file to calculate p95 and p99 durations from durations.csv file

sachadorf13:
### What?
Script to output p95 and p99 durations data from durations.csv

#### Changes:
- Added TypeScript file to output p95 and p99 durations data using generated durations.csv

---

### Why?
To make the process of getting the p95 and p99 durations metrics more efficient and more accurate
Note: This logic can probably be improved on to get all of the metrics in the same step rather than two, but this works for the time being

---


Tom-Dann (commented):
This will need to be setup as a proper project to ensure the dependencies are tracked, it can be transpiled in a reproducible way and linting is applied.

PR #521, opened 2024-03-07
[QA-447] Removing RP-stub check in Auth sign-in journey

maddiemorris:
### What?
- removes the RP Stub check

#### Changes:
- remove rpStubCheck boolean
- edit `B02_SignIn_05_AuthMFA_EnterTOTP POST` and `B02_SignIn_07_SMSMFA_EnterOTP POST` steps, to an `if > else` statement to only check for terms and conditions update, and remove the RP stub checks.
---

### Why?
- to run a successful Authentication signIn journey

### DIFF
--- a/deploy/scripts/src/authentication/test.ts
+++ b/deploy/scripts/src/authentication/test.ts
@@ -307,30 +307,6 @@ export function signUp (): void {
     { isStatusCode200, ...pageContentCheck('User information') })
   })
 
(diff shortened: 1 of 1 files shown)


Tom-Dann:
Whilst we are removing stub only calls, can you remove the logout step from the `signUp` scenario too?

Tom-Dann (approved):

PR #532, opened 2024-03-15
[QA-451] Standardise load profile definitions

Tom-Dann:
### What?
Standardise the way load profiles are defined across all test scripts. Introduces four different standard types of load profiles:
- `smoke`
  - Max VU: `1`
  - Target: `1`
  - Stages: 
    1. `1` minute
- `short`:
  - Stages:
    1.  `5` min ramp up
    2. `15` min steady state
    3. `5` min ramp down
- `full`:
  - Stages:
    1.  `15` min ramp up
    2. `30` min steady state
    3. `5` min ramp down
- `deployment`:
  - Stages:
    1.  `5` min ramp up
    2. `20` min steady state
    3. `5` min ramp down
 
#### Changes:
- `deploy/scripts/src/common/utils/config/load-profiles.ts`: Added a `createScenario()` utility function which creates a scenario given the profile parameters. `LoadProfile` is an enum containing the four different types of tests introduced: `smoke`, `short`, `full` and `deployment`
- `deploy/scripts/src/common/unit-tests.ts`: Added unit tests to validate `createScenario()` is working as expected
- `deploy/scripts/src/**/*.ts`: Updated all test scripts to use these new four load profiles types where applicable. (Also tidied up superfluous naming in these scenarios e.g. `lowVolumeTest` -> `lowVolume`, `addressScenario1` -> `address`)

---

### Why?
The current test scripts have a large amount of duplication of code where we are defining similar load profiles in each script,  also many of these load profiles were slightly inconsistent. This change reduces this duplication and also standardises common load profiles, so that test profiles are consistent across teams.


sandy-may:
Please may we get a Readme update for this?

Tom-Dann:
> Please may we get a Readme update for this?

There is a JSDoc on the new function, which explains how the parameters are used, and has an example of how to use it in a script. I'm keen to keep as much of the documentation of functions as close to where they are actually defined.

Would you perhaps prefer an additional README doc which gives a directory of all the utility functions we have, and how they are used within scripts?

sandy-may:
> > Please may we get a Readme update for this?
> 
> There is a JSDoc on the new function, which explains how the parameters are used, and has an example of how to use it in a script. I'm keen to keep as much of the documentation of functions as close to where they are actually defined.
> 
> Would you perhaps prefer an additional README doc which gives a directory of all the utility functions we have, and how they are used within scripts?

I think that'd be helpful, yes 👍 

I think the litmus test here is: if a new user was to create a new test script, do we have sufficient guidance that they can follow for a script to be created following the standards you're trying to introduce here?

Tom-Dann:
These utility functions are not enforced, more so to save time and code duplication, but I agree that it would be helpful to have a directory of them. Currently for new developers we have the existing scripts; example scripts in `scripts/src/examples`; READMEs in `/` and `/deploy/scripts`; JSDoc documentation on utility functions; and example usage of utility functions in `scripts/src/common/unit-tests.ts`.

I'll add a ticket to automatically generate further documentation using the JSDocs - that way we won't be duplicating documentation over multiple files - but I'll leave that for scope of another another PR. I think the `+300/-1400` is plenty to review in one change for now :)

Tom-Dann:
@sandy-may Documentation to be added in QA-454

Tom-Dann:
@sandy-may After merging #543 and #545, additional documentation for utility functions is available on the [github pages site](https://govuk-one-login.github.io/performance-testing)

PR #564, opened 2024-04-09
QA-472 - Lambda based stub for RP testing.

andyloughran:
### What?
Introduced a new SAM stack and pipleine into the performance environments of a test RP stub.

#### Changes:
- New sam stack 'koa-stub'
---

### Why?
Allow performance tests to scale faster.
---


andyloughran:
Need to fix the template to be compatible with secure pipelines - the serverless connectors currently dont' allow permissions policies to be attached

.github/workflows/koa-stub-main.yaml
```diff
@@ -0,0 +1,90 @@
+name: Kob Stub Deploy
```
Tom-Dann, line 1:
```suggestion
name: Koa Stub Deploy
```

Tom-Dann (commented):
Can you also add dependabot updates for the npm project in `.github/dependabot.yml`?

Tom-Dann (approved):

PR #619, opened 2024-04-30
[QA-496] New Loadprofile for Static Asset Testing

maddiemorris:
### What?
Added a new loadProfile with an extended ramp-up and reduced steady state for static asset testing

#### Changes:
Added a new LoadProfile for static asset testing of:
- Target volume: 30 iterations per second 
- Ramp-up: 15 minutes
- Steady-state: 5 minutes
- Ramp-down: 5 minutes

---

### Why?
To allow for more accurate static asset testing

---

### DIFF
--- a/deploy/scripts/src/common/utils/config/load-profiles.ts
+++ b/deploy/scripts/src/common/utils/config/load-profiles.ts
@@ -94,7 +94,7 @@ export enum LoadProfile {
   short,
   full,
   deployment,
-  extendedRampUp
+  rampOnly
 }
 function createStages(type: LoadProfile, target: number): Stage[] {
   switch (type) {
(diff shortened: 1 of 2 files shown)


deploy/scripts/src/common/utils/config/load-profiles.ts
```diff
@@ -93,7 +93,8 @@ export enum LoadProfile {
   smoke,
   short,
   full,
-  deployment
+  deployment,
+  extendedRampUp
```
Tom-Dann, line 97: Perhaps call rename this to `rampOnly`? This currently makes it seem like it's a longer ramp up than usual which isn't the case

Tom-Dann (approved):

PR #632, opened 2024-05-07
QA-535 Increase timeout of openid-client

andyloughran:
### What?
Increase the timeout of the RP calling auth, as it times out after 3500ms by default

#### Changes:
- Double the default timeout to 7000ms

---

### Why?
Avoid the RP itself being the bottleneck.


Tom-Dann (approved):

paroxp (approved):
This will setup the timeout globally, right? ie. for the process?

Would we not rather set the timeout on the client?

PR #656, opened 2024-05-22
[QA-510]: Updated body for persist VC call

abhinaysrivastavaperf:
### What?
Updated body for persist VC call

#### Changes:
- Updated body for persist VC call based on API spec

---

### Why?
We are getting HTTP-400 requests for persist VC calls because of the incorrect payload.


deploy/scripts/src/accounts/vc-storage.ts
```diff
@@ -140,3 +142,27 @@ function getToken(r: Response): string {
   if (token !== null && typeof token === 'string') return token
   fail('token not found')
 }
+
+interface vcPayload {
```
Tom-Dann, line 146:
```suggestion
interface VCPayload {
```
Fix the SonarCloud code smell

abhinaysrivastavaperf: Fixed now

Tom-Dann (approved):

PR #657, opened 2024-05-23
[QA-510]: Add other VCs in the persist VC call and modify payload

abhinaysrivastavaperf:
### What?
Updated persist VC request body to include other payloads

#### Changes:
- Modified the body of the persist VC call as `metadata` is not currently included in the current version of code in staging.  
- Added `passport` and `kbv` VCs in the body of the persist VC call.

---

### Why?
To fix the `HTTP-400` Bad Request errors for persist VC call in staging.


deploy/scripts/src/accounts/vc-storage.ts
```diff
@@ -89,10 +89,19 @@ export function persistVC(): void {
...
   }
 
-  const body = JSON.stringify(generateVCPayload(jwtTokens.fraud))
+  // const body = JSON.stringify(generateVCPayload(jwtTokens.fraud))
+
+  const body = JSON.stringify([
+    generateVCPayload(jwtTokens.fraud),
+    generateVCPayload(jwtTokens.passport),
+    generateVCPayload(jwtTokens.kbv)
+  ])
```
Tom-Dann, line 104:
```suggestion
  const body = JSON.stringify(Object.values(jwtTokens).map(generateVCPayload))
```

deploy/scripts/src/accounts/vc-storage.ts
```diff
@@ -89,10 +89,19 @@ export function persistVC(): void {
...
   }
 
-  const body = JSON.stringify(generateVCPayload(jwtTokens.fraud))
+  // const body = JSON.stringify(generateVCPayload(jwtTokens.fraud))
```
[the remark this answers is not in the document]
abhinaysrivastavaperf: Removed

deploy/scripts/src/accounts/vc-storage.ts
abhinaysrivastavaperf: Updated

Tom-Dann (approved):

PR #693, opened 2024-06-18
QA-516 Script for Fraud testing

sachadorf13:
### What?
Create a script for fraud testing

#### Changes:
- Added a step call Cognito to retrieve the access token
- Added a step to send pre-signed event to SSF
- Now using pre-signed payload using a separate script provided by fraud team
- Moved getAccessToken function into common directory and updated Kiwi & fraud files to use this

---

### Why?
To start performance testing SSF

---

### Related:
- [Fraud Performance Testing](https://govukverify.atlassian.net/wiki/spaces/FPAD/pages/4224385577/Performance+Testing)


deploy/scripts/src/fraud/test.ts
```diff
@@ -0,0 +1,93 @@
+import { type Options } from 'k6/options'
+import {
+  selectProfile,
+  type ProfileList,
+  describeProfile,
+  createScenario,
+  LoadProfile
+} from '../common/utils/config/load-profiles'
+import { getAccessToken } from '../cri-kiwi/utils/authorization'
```
Tom-Dann, line 9: If you need this function it should probably be moved to a common folder, as it's no longer just applicable to kiwi

Tom-Dann (approved):

PR #797, opened 2024-08-13
[QA-550] Add debug logging to OLH changeEmail

maddiemorris:
### What?
Add debug logging to the OLH `changeEmail` test script

#### Changes:
- Add and `if` statement where if the `pageContentCheck` fails, the unexpected page content will be returned.

---

### Why?
- If we get response validation failures, we will be able to see the content of the incorrect page returned to help with debugging.

### DIFF
--- a/deploy/scripts/src/accounts/test.ts
+++ b/deploy/scripts/src/accounts/test.ts
@@ -227,7 +227,7 @@ export function changeEmail(): void {
     () => {
       const r = http.get(env.envURL + '/security')
       if (!pageContentCheck('Delete your GOV.UK One Login').validatePageContent(r)) {
-        console.log(' Expected "Delete your GOV.UL One Login", got: ', r.html('h2').eq(3).text())
+        console.log(' Expected "Delete your GOV.UK One Login", got: ', r.html('h2').eq(3).text())
       }
       return r
     },
--- a/deploy/scripts/src/accounts/test.ts
+++ b/deploy/scripts/src/accounts/test.ts
@@ -179,7 +179,7 @@ export function changeEmail(): void {
       () => {
         const r = http.get(res.headers.Location)
         if (!pageContentCheck('API Simulation Tool').validatePageContent(r)) {
-          console.log(r.html('h1').first().text())
+          console.log('Expected "API Simulation Tool", got: ', r.html('h1').first().text())
         }
         return r
       },
@@ -207,8 +207,8 @@ export function changeEmail(): void {
       groups[5].split('::')[1],
       () => {
         const r = http.get(res.headers.Location)
-        if (!pageContentCheck('Services you can use with GOV.UK One Login--').validatePageContent(r)) {
-          console.log(r.html('h2').eq(1).text())
+        if (!pageContentCheck('Services you can use with GOV.UK One Login').validatePageContent(r)) {
+          console.log(' Expected "Services you can use with GOV.UK One Login", got: ', r.html('h2').eq(1).text())
         }
         return r
       },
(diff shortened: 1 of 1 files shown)


deploy/scripts/src/accounts/test.ts
```diff
@@ -178,6 +178,9 @@ export function changeEmail(): void {
       isStatusCode200,
       ...pageContentCheck('API Simulation Tool')
     })
+    if (!pageContentCheck('API Simulation Tool').validatePageContent(res)) {
```
Tom-Dann, line 181: I think this will never be called as this check needs to happen before the `timeGroup` checks which call `fail` if the checks fail. 
Like this
```suggestion
      res = timeGroup(groups[2].split('::')[1], () => {
      const r = http.get(res.headers.Location)
      if (!pageContentCheck('API Simulation Tool').validatePageContent(r)) {
        console.log(r.html('h1').first().text())
      }
      return r
    }, {
      isStatusCode200,
      ...pageContentCheck('API Simulation Tool')
    })
    
```

Tom-Dann (changes requested):

deploy/scripts/src/accounts/test.ts
```diff
@@ -174,10 +174,20 @@ export function changeEmail(): void {
...
     //02_OIDCStubCall
-    res = timeGroup(groups[2].split('::')[1], () => http.get(res.headers.Location), {
-      isStatusCode200,
-      ...pageContentCheck('API Simulation Tool')
-    })
+    res = timeGroup(
+      groups[2].split('::')[1],
+      () => {
+        const r = http.get(res.headers.Location)
+        if (!pageContentCheck('--API Simulation Tool--').validatePageContent(r)) {
```
Tom-Dann, line 181:
```suggestion
        if (!pageContentCheck('API Simulation Tool').validatePageContent(r)) {
```
I think these include `-` for debugging, remove these so they are the same as the assertion checks

deploy/scripts/src/accounts/test.ts
```diff
@@ -174,10 +174,20 @@ export function changeEmail(): void {
...
+          console.log(r.html('h1').first().text())
```
Tom-Dann, line 182:
```suggestion
          console.log(groups[2].split('::')[1], r.html('h1').first().text())
```
Suggest adding some info to the start of the console logs, so that we can tell what the expected result is meant to be

Tom-Dann: Or 
```suggestion
          console.log('Expected "API Simulation Tool", got: ',r.html('h1').first().text())
```

Tom-Dann (changes requested):

deploy/scripts/src/accounts/test.ts
```diff
@@ -193,72 +203,123 @@ export function changeEmail(): void {
...
-  res = timeGroup(groups[6], () => http.get(env.envURL + '/security'), {
-    isStatusCode200,
-    ...pageContentCheck('Delete your GOV.UK One Login')
-  })
+  res = timeGroup(
+    groups[6],
+    () => {
+      const r = http.get(env.envURL + '/security')
+      if (!pageContentCheck('Delete your GOV.UK One Login').validatePageContent(r)) {
+        console.log(' Expected "Delete your GOV.UL One Login", got: ', r.html('h2').eq(3).text())
```
Tom-Dann, line 230:
```suggestion
        console.log(' Expected "Delete your GOV.UK One Login", got: ', r.html('h2').eq(3).text())
```

Tom-Dann (approved):

PR #800, opened 2024-08-16
QA-553 Ensure unique user data for each test

andyloughran:
### What?
Ensure's that the uniqueness of virtualUser with userId.

#### Changes:
Follows guidelines: https://community.grafana.com/t/unique-login-for-each-virtual-user/99596 

---

### Why?
In the auth test, the logs showed the same user was passing through the journey concurrently, causing an increase in errors.  This change (to derive from vu) ensures that no vu has a duplicate id when running concurrently through the system.

---

### Related:

### DIFF
--- a/deploy/scripts/src/authentication/test.ts
+++ b/deploy/scripts/src/authentication/test.ts
@@ -17,6 +17,7 @@ import { iterationsCompleted, iterationsStarted } from '../common/utils/custom_m
 import { timeGroup } from '../common/utils/request/timing'
 import { getEnv } from '../common/utils/config/environment-variables'
 import { browser, type Page, type Response as PageResponse } from 'k6/browser'
+import { vu } from 'k6/execution';
 
 const profiles: ProfileList = {
   smoke: {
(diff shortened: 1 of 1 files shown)


Tom-Dann (changes requested):
You'll need to update the load profiles to ensure that the number of virtual users is capped at 10,000

```diff
  load20: {
-    ...createScenario('signIn', LoadProfile.full, 380),
+    ...createScenario('signIn', LoadProfile.full, 380, 26),
    ...createScenario('signUp', LoadProfile.full, 20)
  },
  load: {
-    ...createScenario('signIn', LoadProfile.full, 500)
+    ...createScenario('signIn', LoadProfile.full, 500, 20)
  },
  stress: {
-    ...createScenario('signIn', LoadProfile.full, 2000),
+    ...createScenario('signIn', LoadProfile.full, 2000, 5),
    ...createScenario('signUp', LoadProfile.full, 100)
  },
```

Tom-Dann (approved):

PR #822, opened 2024-08-30
[QA-561] Removing the sign-out step in the OLH changeEmail scenario

maddiemorris:
### What?
Removed the sign-out step from the OLH `changeEmail` scenario

#### Changes:
Removes `B01_ChangeEmail_08_Signout` from the `changeEmail` scenario

---

### Why?
To run a successful OLH `changeEmail` test, as currently when one user signs-out, all other concurrent users sign-out, resulting in 99% user journey failure.


abhinaysrivastavaperf (changes requested):
You can remove the sign out step from the other scenarios as well if this theory is applicable to those scenarios.

maddiemorris:
> You can remove the sign out step from the other scenarios as well if this theory is applicable to those scenarios

I have removed it from all but `deleteAccount` as signing-out is necessary to delete the account in the last step

abhinaysrivastavaperf (approved):

PR #827, opened 2024-09-02
OJ-2748 lambda warmer stage added to performance tests

sandragds:
### What?
Set up stage for imposter stub to warm up lambdas before performance testing is ran.

#### Changes:
- function added to run requests to imposter stub in nino-check
- nino-check.ts script modified to include the request before the performance test
---

### Why?

- Due to lambda cold start issue, performance tests are slow in imposter stubs. To mediate the issue, lambdas can be warmed up by sending requests to the imposter stubs before the performance tests.  
---

### Related:
- [OJ-2748](https://govukverify.atlassian.net/browse/OJ-2748)

[OJ-2748]: https://govukverify.atlassian.net/browse/OJ-2748?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


abhinaysrivastavaperf:
Just to highlight, the latest changes mean that the lambda warmer call will be executed each iteration.

deploy/scripts/src/cri-orange/nino-check.ts
```diff
@@ -83,7 +82,18 @@ const csvData1: Nino[] = new SharedArray('csvDataNino', () => {
...
+}
+
 export function ninoCheck(): void {
+  let imposterstarted = false
+  if (!imposterstarted) {
+    const response = imposter.handler()
+    console.log('Response status: ', response.status)
+    console.log('Response body: ', response.body)
+    imposterstarted = true
+  }
```
Tom-Dann, line 96:
```suggestion
    const response = imposter.handler()
    console.log('Response status: ', response.status)
    console.log('Response body: ', response.body)
```
This code is functionally the same as removing the `imposterstarted` variable and the if statement, as it will always run.

If you are trying to run this once at the beginning of the test I'd move this call to the `setup()` function, or if you want to run it once per VU the `imposterstarted` will need to be declared outside of the function scope.

Tom-Dann (changes requested):

deploy/scripts/src/cri-orange/nino-check.ts
abhinaysrivastavaperf: Yea I included this in the `setup()` function initially but I had a chat with Sandra and was advised that this needs to run periodically instead of only once. 
To avoid running this as many time as the actual test iterations, we may have a separate function for this which will be called like 10% or 20% of the total number of executed iterations

Tom-Dann: Okay, well that approach will need to be implemented and the code smell fixed first before merging

abhinaysrivastavaperf: Updated the approach to make 10 concurrent request (as a batch request) in the `setUp` function and fixed the code smell.

deploy/scripts/src/cri-orange/nino-check.ts
```diff
@@ -83,6 +83,16 @@ const csvData1: Nino[] = new SharedArray('csvDataNino', () => {
...
 })
 
+export function setup(): void {
+  describeProfile(loadProfile)
+  const imposterstarted = false
+  if (!imposterstarted) {
+    const response = imposter.handler()
+    console.log('Response status: ', response[0].status)
+    console.log('Response body: ', response[0].body)
+  }
```
Tom-Dann, line 93:
```suggestion
  const response = imposter.handler()
  console.log('Response status: ', response[0].status)
  console.log('Response body: ', response[0].body)
```
Remove the useless variable assignment and if condition

Tom-Dann (changes requested):

deploy/scripts/src/cri-orange/nino-check.ts
abhinaysrivastavaperf: Done

Tom-Dann (approved):
Update the PR title, but otherwise okay

abhinaysrivastavaperf:
Closing as this is not required anaymore.

PR #828, opened 2024-09-03
[QA-564] Adding the logout step back into OLH scenarios.

maddiemorris:
### What?
Adds the logout step back into the OLH scenarios 

#### Changes:
- Adds the `B01_ChangeEmail_08_Logout`, `B02_ChangePassword_07_SignOut`, and `B03_ChangePhone_08_SignOut`steps back into the `changeEmail`, `changePassword`, and `changePhone` scenarios of the OLH test script.

---

### Why?
The OIDC stub has been updated to use unique session IDs for each user so only the singular user will be logged out. This should provide a fix for the response validation errors while still allowing the sessions to be deleted from the Session store during the logout step.


abhinaysrivastavaperf (approved):
Could have just reverted this PR instead of raising a new one but either way is fine
https://github.com/govuk-one-login/performance-testing/pull/822/files

PR #855, opened 2024-09-18
[QA-576]: IPV Core: Updated script based on latest changes in user journey

abhinaysrivastavaperf:
### What?
Updated IPV core passport script based on latest changes in user journey

#### Changes:
- Updated the page content after the DCMAW stub step.
- Updated CloudFormation template with path for env variable. The respective parameter has been created in SSM.

---

### Why?
To fix the IPV core script.

### DIFF
--- a/deploy/scripts/src/ipv-core/test.ts
+++ b/deploy/scripts/src/ipv-core/test.ts
@@ -189,7 +189,7 @@ if (!validEnvironments.includes(environment))
 const env = {
   orchStubEndPoint: __ENV[`IDENTITY_${environment}_ORCH_STUB_URL`],
   ipvCoreURL: __ENV[`IDENTITY_${environment}_CORE_URL`],
-  vtrText: __ENV[`IDENTITY_${environment}_VTR_TEXT`]
+  vtrText: __ENV[`IDENTITY_${environment}_CORE_VTR_TEXT`]
   // staticResources: getEnv('K6_NO_STATIC_RESOURCES') !== 'true'
 }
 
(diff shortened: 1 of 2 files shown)


deploy/scripts/src/ipv-core/test.ts
```diff
@@ -189,7 +189,7 @@ if (!validEnvironments.includes(environment))
 const env = {
   orchStubEndPoint: __ENV[`IDENTITY_${environment}_ORCH_STUB_URL`],
   ipvCoreURL: __ENV[`IDENTITY_${environment}_CORE_URL`],
-  vtrText: __ENV[`IDENTITY_${environment}_VTR_TEXT`]
+  vtrText: __ENV[`IDENTITY_${environment}_CORE_VTR_TEXT`]
```
Tom-Dann, line 192: This new variable is not being used anywhere. Is there a matching line missing?

abhinaysrivastavaperf: This is not a new variable. I have just updated the name of the env var to keep it in line with the naming convention that we are following. The object property `env.vtrText` is being used in Line 233

``/authorize?journeyType=full&userIdText=${userId}&signInJourneyIdText=${signInJourneyId}&vtrText=${env.vtrText}&targetEnvironment=${environment}&reproveIdentity=NOT_PRESENT&emailAddress=${testEmail}&votText=&jsonPayload=&evidenceJsonPayload=&error=recoverable`,`

Tom-Dann: Right you are

Tom-Dann (approved):

PR #865, opened 2024-09-26
[QA-579]: Update Auth scripts based on latest changes.

abhinaysrivastavaperf:
### What?
Update Auth scripts based on latest changes

#### Changes:
- Updated signUp and signIn scenarios to use the new orch stub.

---

### Why?
Auth user journeys have a new stub as the entry point as part of latest changes. The script changes are to reflect these changes in the user journeys.

### DIFF
--- a/deploy/scripts/src/authentication/test.ts
+++ b/deploy/scripts/src/authentication/test.ts
@@ -470,7 +470,7 @@ export function orchStubSubmit(groups: readonly string[]): Response {
 export function signIn(): void {
   let res: Response
   const groups = groupMap.signIn
-  const userData = dataSignIn[execution.scenario.iterationInInstance % dataSignIn.length]
+  const userData = dataSignIn[execution.vu.idInTest - 1]
   iterationsStarted.add(1)
 
   // B02_SignIn_01_OrchStubSubmit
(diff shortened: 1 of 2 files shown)


deploy/scripts/src/authentication/test.ts
```diff
@@ -443,18 +417,69 @@ export function initializeJourney(groups: readonly string[]): Response {
   })
 }
 
+export function orchStubSubmit(groups: readonly string[]): Response {
```
Tom-Dann, line 420: Is this function replacing the `initializeJourney()` function?

abhinaysrivastavaperf: Yes

Tom-Dann: Okay I suggest removing that function as part of this PR in that case

abhinaysrivastavaperf: Done

deploy/scripts/src/authentication/test.ts
```diff
@@ -567,25 +534,13 @@ export function signIn(): void {
...
       })
     })
   }
 
-  // 25% of users logout
-  if (Math.random() <= 0.25) {
-    sleep(1)
-
-    // B02_SignIn_07_Logout
-    logout(groups.slice(17, 21))
```
Tom-Dann, line 588: Can remove the `logout()` function too if it is no longer being used

Tom-Dann (changes requested):

deploy/scripts/src/authentication/test.ts
abhinaysrivastavaperf: Removed

PR #1032, opened 2025-01-20
[QA-726]: Adhoc address scenario with consistent virtual user count

abhinaysrivastavaperf:
### What?
Added an adhoc scenario for Address to test with consistent virtual user count.

#### Changes:
- Added the function `addressAdhocScenario()`.
- Removed `sleep()` after all the steps in the above scenario to evaluate max support performance.
- Removed the final stub call to the `/callback` endpoint.

---

### Why?
To conduct performance test of Address CRI under a load of 10 concurrent virtual users.


andyloughran (approved):
This looks good to me, based on VUs will be more helpful

maddiemorris (approved):

PR #1035, opened 2025-01-22
[QA-744]: Add a separate profile for Address scenario

abhinaysrivastavaperf:
### What?
Add a separate profile for Address scenario

#### Changes:
- Added a separate load profile for Address scenario with the `ramping-arrival-rate` executor with the `timeUnit` as **10 seconds**. Adjusted the `target` load accordingly and updated the `preAllocatedVU` and `maxVUs` count to match baseline.
- Updated the `timeRequest` function to return the HTTP error status and body if an assertion fails.

---

### Why?
To conduct a performance test for Address scenario with the updated config.

### DIFF
--- a/deploy/scripts/src/common/utils/request/timing.ts
+++ b/deploy/scripts/src/common/utils/request/timing.ts
@@ -41,7 +41,7 @@ export function timeRequest<T>(fn: () => T, checks: Checkers<T> = {}): T {
   if (check(res, checks)) {
     durations.add(duration)
   } else {
-    fail('Response validation failed')
+    fail(res.status + res.body)
   }
   return res
 }
--- a/deploy/scripts/src/cri-orange/address.ts
+++ b/deploy/scripts/src/cri-orange/address.ts
@@ -64,6 +64,21 @@ const profiles: ProfileList = {
       ],
       exec: 'addressAdhocScenario'
     }
+  },
+  addressUpdatedConfig: {
+    address: {
+      executor: 'ramping-arrival-rate',
+      startRate: 1,
+      timeUnit: '10s',
+      preAllocatedVUs: 100,
+      maxVUs: 400,
+      stages: [
+        { target: 130, duration: '129s' },
+        { target: 130, duration: '900s' },
+        { target: 0, duration: '300s' }
+      ],
+      exec: 'address'
+    }
   }
 }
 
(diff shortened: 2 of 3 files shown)


deploy/scripts/src/common/utils/request/timing.ts
```diff
@@ -41,7 +41,7 @@ export function timeRequest<T>(fn: () => T, checks: Checkers<T> = {}): T {
   if (check(res, checks)) {
     durations.add(duration)
   } else {
-    fail('Response validation failed')
+    fail(res.status + res.body)
```
andyloughran, line 44: Long term let's make this configurable via local env var.

deploy/scripts/src/cri-orange/address.ts
```diff
@@ -64,6 +64,21 @@ const profiles: ProfileList = {
...
+  addressUpdatedConfig: {
+    address: {
+      executor: 'ramping-arrival-rate',
+      startRate: 1,
+      timeUnit: '10s',
+      preAllocatedVUs: 100,
+      maxVUs: 400,
+      stages: [
+        { target: 130, duration: '129s' },
+        { target: 130, duration: '900s' },
```
andyloughran, line 77: Let's reduce this time for peak to a shorter length of time - e.g. 120s

deploy/scripts/src/cri-orange/address.ts
```diff
@@ -64,6 +64,21 @@ const profiles: ProfileList = {
...
+        { target: 0, duration: '300s' }
```
andyloughran, line 78: I think we can shorten this time significantly - we don't need this to pass any of our NFRs - this stage feels superfluous

deploy/scripts/src/cri-orange/address.ts
abhinaysrivastavaperf: We are going to compare the results of this test against the baseline test for address which had a steady state duration of 15 minutes. Suggest we keep this to 15 minutes for the initial run and reduce it once we have a view of the results.

deploy/scripts/src/cri-orange/address.ts
abhinaysrivastavaperf: This was to allow the load to ramp down from 13 to 0 in 5 minutes instead of a sudden ramp down. Agree that this is not validating any NFRs so have reduced this to 2 minutes.

andyloughran (approved):

PR #1044, opened 2025-01-30
[QA-750]: Updated startRate and timeUnit in Address CRI

abhinaysrivastavaperf:
### What?
Updated startRate and timeUnit in Address CRI

#### Changes:
- Updated startRate to 6 and timeUnit to 1 minute in Address CRI

---

### Why?
To validate that the performance of the Address CRI is similar with the updated test design.


richie-co (approved):
Approved

PR #1056, opened 2025-02-07
[QA-762] Updating OLH and Auth Load Profiles

maddiemorris:
### What?
Updates the OLH and Auth scenarios with the new `lowVolPerf007Test` profile.

#### Changes:
- Adds the `lowVolPerf007Test` load profile to OLH `changeEmail`, `changePassword`, `changePhone`, and `deleteAccount` scenarios at a target volume of 2j/s, and to the Auth `signIn` and `signUp` at 10j/s and 2j/s respectively. 

---

### Why?
To run flat load profile tests for OLH and Auth


abhinaysrivastavaperf (approved):
We will need to get a confirmation on the OLH volumes before running the test at this volume.

PR #1078, opened 2025-02-17
[QA-773] - Parameterise Load Profile for Pipeline Execution to eliminate Hardcoded Profiles

umairfarooquigds:
### What?
This Pull Request introduces parameterisation to our k6 performance test scripts, enabling us to dynamically configure load profiles without code modifications.

#### Changes:

- Environment Variable based Load Profile: The k6 script now reads key load profile settings from environment variables:
- K6_START_RATE: Initial arrival rate of virtual users.
- K6_TIME_UNIT: Time unit for arrival rate calculations
- K6_PRE_ALLOCATED_VUS: Number of pre-allocated virtual users.
- K6_MAXVUS: Maximum concurrent virtual users.
- K6_RU_TARGET: Target arrival rate for the ramp-up phase.
- K6_RU_DURATION: Duration of the ramp-up phase (in seconds).
- K6_SS_TARGET: Target arrival rate for the steady-state phase.
- K6_SS_DURATION: Duration of the steady-state phase (in seconds).

---

### Why?
We can now quickly adjust load test parameters without modifying the script, speeding up testing cycles
---

### DIFF
--- a/deploy/scripts/src/olh/test.ts
+++ b/deploy/scripts/src/olh/test.ts
@@ -18,6 +18,19 @@ import { sleepBetween } from '../common/utils/sleep/sleepBetween'
 import { getEnv } from '../common/utils/config/environment-variables'
 import { getThresholds } from '../common/utils/config/thresholds'
 
+const dprofile = {
+  // scenarioName: getEnv('K6_SCENARIO_NAME'),
+  // executor: getEnv('K6_EXECUTOR'),
+  startRate: parseInt(getEnv('K6_START_RATE') || '1', 10),
+  timeUnit: getEnv('K6_TIME_UNIT'),
+  preAllocatedVUs: parseInt(getEnv('K6_PRE_ALLOCATED_VUS') || '100', 10),
+  maxVUs: parseInt(getEnv('K6_MAXVUS') || '400', 10),
+  rutarget: parseInt(getEnv('K6_RU_TARGET') || '20', 10),
+  ruduration: getEnv('K6_RU_DURATION'),
+  sstarget: parseInt(getEnv('K6_SS_TARGET') || '20', 10),
+  ssduration: getEnv('K6_SS_DURATION')
+}
+
 const profiles: ProfileList = {
   smoke: {
     ...createScenario('changeEmail', LoadProfile.smoke),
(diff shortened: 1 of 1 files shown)


deploy/scripts/src/olh/test.ts
```diff
@@ -18,6 +18,19 @@ import { sleepBetween } from '../common/utils/sleep/sleepBetween'
 import { getEnv } from '../common/utils/config/environment-variables'
 import { getThresholds } from '../common/utils/config/thresholds'
 
+const dprofile = {
+  // scenarioName: getEnv('K6_SCENARIO_NAME'),
```
abhinaysrivastavaperf, line 22: Commented variables can be removed

deploy/scripts/src/olh/test.ts
```diff
@@ -18,6 +18,19 @@ import { sleepBetween } from '../common/utils/sleep/sleepBetween'
...
+  // executor: getEnv('K6_EXECUTOR'),
```
abhinaysrivastavaperf, line 23: Can be removed

deploy/scripts/src/olh/test.ts
```diff
@@ -18,6 +18,19 @@ import { sleepBetween } from '../common/utils/sleep/sleepBetween'
...
+  startRate: parseInt(getEnv('K6_START_RATE') || '1', 10),
+  timeUnit: getEnv('K6_TIME_UNIT'),
+  preAllocatedVUs: parseInt(getEnv('K6_PRE_ALLOCATED_VUS') || '100', 10),
+  maxVUs: parseInt(getEnv('K6_MAXVUS') || '400', 10),
+  rutarget: parseInt(getEnv('K6_RU_TARGET') || '20', 10),
```
abhinaysrivastavaperf, line 28: The target for both ramp up and steady state phases will be the same. So we can just keep one variable

deploy/scripts/src/olh/test.ts
```diff
@@ -84,6 +97,22 @@ const profiles: ProfileList = {
       ],
       exec: 'deleteAccount'
     }
+  },
+  dynamicProfile: {
+    changeEmail: {
+      //executor: dprofile.executor,
```
abhinaysrivastavaperf, line 103: Remove commented variable

abhinaysrivastavaperf (changes requested):

deploy/scripts/src/olh/test.ts
umairfarooquigds: Removed

deploy/scripts/src/olh/test.ts
umairfarooquigds: Removed

deploy/scripts/src/olh/test.ts
umairfarooquigds: updated with variable name as 'target'.

abhinaysrivastavaperf (changes requested):
How can we run multiple scenarios of the same script using this approach?

abhinaysrivastavaperf:
> > How can we run multiple scenarios of the same script using this approach?
> > I think we will have to have multiple environment variables in the dynamic load profiles specific to each scenario e.g. K6_START_RATE_IDENTITY and K6_START_RATE_IDREUSE etc.

Well in that case, the only benefit this PR gives is that we can run the test in the pipeline but from a code repeatability perspective, we still need to optimise this further. Let us explore this further. For today's test, we can add hard coded profiles which will be replaced once this is sorted.

PR #1087, opened 2025-02-21
[QA-779] Reducing the Fraud SSF Inbound target volume

maddiemorris:
### What?
Reduces the Fraud SSF Inbound target volume to 3 messages per second

#### Changes:
Reduces the Fraud SSF Inbound target volume to 3 messages per second

---

### Why?
To run a performance test for Fraud SSF Inbound at 3 messages per second.


abhinaysrivastavaperf (approved):

abhinaysrivastavaperf:
We can use the short load profile as we do not need a 15 minute ramp up

PR #1209, opened 2025-04-08
QA-867 - Adding Data generation Shell Script 

NarasimhaYanamala:
### What?
Data Generation for AIS, IPV Core EVCS. Separate data generation  shell script.

#### Changes:
At Present, we are replacing Data Gen Shell script from Confluence page to the existing Shell Script and Once data is generated unsaying the file to keep the old one.  It has a risk of losing exiting script. So Adding Seprate Shell script for Data Generation.

Impacted CRI’s in PT script for Data generation:

IPV Core

AIS

EVCS

---

### Why?
Instead of replacing the code in existing shell script adding dedicated one for the ease of use.

---

### DIFF
--- a/deploy/reporting/DataGeneration.sh
+++ b/deploy/reporting/DataGeneration.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+# Utility script for reporting on key metrics from a k6 performance test using
+# the JSON file output
+#
+# ARGUMENTS:
+#  $1: path to a .gz archive containing test results in JSON format
+#  $2: folder to use a working and output directory
+set -e
+INPUT=${1:-TestData.gz}
+OUTPUT=${2:-results_$(date +%s)}
+mkdir -p $OUTPUT
+if [[ ! -f "$OUTPUT/TestData.json" ]]; then # Do not unzip or copy if this has already been done
+    if [[ $INPUT == *.gz ]]; then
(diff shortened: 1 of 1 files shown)


deploy/reporting/DataGeneration.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+# Utility script for reporting on key metrics from a k6 performance test using
```
abhinaysrivastavaperf, line 2: Can you update the description here to reflect the correct intention of this PR?

deploy/reporting/DataGeneration.sh
```diff
@@ -0,0 +1,34 @@
...
+    if [[ $INPUT == *.gz ]]; then
+        gunzip -c $INPUT >$OUTPUT/TestData.json
+    else
+        cp $INPUT $OUTPUT/TestData.json
+    fi
+fi
+cd $OUTPUT
+echo ''
+echo '┌-----------------------------------------┐'
+echo '| Request Duration by Timestamp and Group |'
```
abhinaysrivastavaperf, line 22: This is Irrelevant as we are not extracting request duration here. Can you update the output statement correctly?

abhinaysrivastavaperf (changes requested):

PR #1216, opened 2025-04-10
[QA-826] Adding Iteration 2 load profiles to IPVR

maddiemorris:
### What?
Adds Iteration 2 peak and spike load profiles for IPVR.

#### Changes:
- Adds the `perf006Iteration2PeakTest` and `spikeI2HighTraffic` load profiles to `cri-kiwi/ipvr.ts` with volumes of:
- peak test: `authEvent` - 11j/s, `allEvents` - 0.3j/s
- spike test: `authEvent` - 32j/s, `allEvents` - 1j/s

---

### Why?
To run iteration 2 peak and spike tests for IPVR

### DIFF
--- a/deploy/scripts/src/cri-kiwi/ipvr.ts
+++ b/deploy/scripts/src/cri-kiwi/ipvr.ts
@@ -30,6 +30,36 @@ const profiles: ProfileList = {
   stress: {
     ...createScenario('authEvent', LoadProfile.full, 2000, 2),
     ...createScenario('allEvents', LoadProfile.full, 12, 2)
+  },
+  perf006Iteration2PeakTest: {
+    authEvent: {
+      executor: 'ramping-arrival-rate',
+      startRate: 2,
+      timeUnit: '1s',
+      preAllocatedVUs: 100,
+      maxVUs: 1000,
+      stages: [
+        { target: 11, duration: '6s' },
+        { target: 11, duration: '30m' }
+      ],
+      exec: 'authEvent'
+    },
+    allEvents: {
+      executor: 'ramping-arrival-rate',
+      startRate: 1,
+      timeUnit: '10',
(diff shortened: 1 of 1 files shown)


deploy/scripts/src/cri-kiwi/ipvr.ts
```diff
@@ -30,6 +30,36 @@ const profiles: ProfileList = {
...
+      stages: [
+        { target: 3, duration: '4s' },
+        { target: 3, duration: '30m' }
+      ],
+      exec: 'allEvents'
+    }
+  },
+  spikeI2HighTraffic: {
+    ...createScenario('authEvent', LoadProfile.spikeI2HighTraffic, 32),
+    ...createScenario('allEvents', LoadProfile.spikeI2HighTraffic, 1)
```
abhinaysrivastavaperf, line 62: Just to clarify - Are we using `spikeI2HighTraffic` because we do not care about the ramp up rate in this test and we know it is going to breach PERF008?

abhinaysrivastavaperf (approved):

deploy/scripts/src/cri-kiwi/ipvr.ts
maddiemorris: @abhinaysrivastavaperf We're using `spikeI2HighTraffic`, as for Iteration 2 we either used to high traffic spike for products that received some element of constant traffic, or low traffic for a CRI that didn't experience constant traffic. I don't think either of them took PERF008 into account. 

We've chosen high traffic for this one as `authEvent` will receive some traffic all the time, and as we wanted to keep both scenarios on the same load profile we chose the same from `allEvents`.

abhinaysrivastavaperf: Thank you

PR #1224, opened 2025-04-14
QA-908 : Lime(Fraud) load profile updates/correction with right Pre & Max vusers

vineeth-kunju:
### What?
Lime(Fraud) load profile updates/correction with right Pre & Max vusers

#### Changes:
The PreAllocated & Max vusers were incorrectly mentioned during the last test and hence couldnt achieve the targeted TPH. This PR is raised to update the preAllocatedVUs to 48 & maxVUs to 96

### Why?
Pre & Max vusers were incorrect to reach the throughput of 16 iterations per second

### DIFF
--- a/deploy/scripts/src/cri-lime/test.ts
+++ b/deploy/scripts/src/cri-lime/test.ts
@@ -275,8 +275,8 @@ const profiles: ProfileList = {
       executor: 'ramping-arrival-rate',
       startRate: 1,
       timeUnit: '10s',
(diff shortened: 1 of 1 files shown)


abhinaysrivastavaperf (changes requested):
Can you update the PR description to reflect a summary of what is being changed? We do not need to add the entire code change in the PR description section.

abhinaysrivastavaperf (approved):

PR #1233, opened 2025-04-15
QA-917:Address(Orange CRI):Perf006-I3 Peak test load profile update

vineeth-kunju:
### What?
Address CRI load profile changes for Perf006 I3 

---

#### Changes:
Address CRI load profile will be updated to simulate 16 iters/sec. The test will run for 30 mins, after the rampup in 161s

---

### Why?
To peak test Address CRI against I3 volume

---


abhinaysrivastavaperf (changes requested):
As discussed, the target for this scenario should be limited to 100 and the ramp duration needs to be updated to 101s
Also the preAllocatedVUs and maxVUs need to be updated as per the revised target.

abhinaysrivastavaperf (approved):

PR #1247, opened 2025-04-17
DCMAW-12496: STS performance tests

jamielutton-gds:
### What?
Add an overview of what this pull request changes

- Adds `getServiceAccessToken.ts` script for getting a service token from STS. This simulates a user logging in to the 'V2 app' and getting an access token that can be used to resume the ID Check journey.

##### `getServiceAccessToken.ts`
- Script for logging in via STS and getting a "service access token" for a protected resource. This simulates the STS part of the journey for completing an ID Check in the 'V2 app'. 
- Notes:
  - The scope of the access token received at the end of this journey is `sts-test.hello-world.read`, because STS does not integrate with ID Check in lower environments. However, the scope of the token does not change the fact that this test still hits all of the APIs that would be hit in an ID Check journey. 
  - There are several calls to STS's /GET .well-known/jwks.json in this test. This is to simulate the different parties (Orchestration, the V2 app and ID Check) retrieving STS's JWKS to verify various tokens.

---

### Why?

Adds scripts for STS performance tests.

---

### Related:


deploy/scripts/src/mobile/sts/utils/assertions.ts
```diff
@@ -0,0 +1,32 @@
...
+  return res.json('client_attestation') !== undefined
+}
+
+export function validateAccessTokenResponse(res: Response): boolean {
+  return res.json('access_token') !== undefined && res.json('id_token') !== undefined
+}
+
+export function validateTokenExchangeResponse(res: Response): boolean {
+  return res.json('access_token') !== undefined
+}
```
jamielutton-gds, line 32: Question: Is it typical to have assertions on the response body, if it is implied that the required fields will be present based on the response being a 200? For example, if any of the above calls return a 200 then by the STS specification there will be an access_token in the response. So it is perhaps redundant to do this check, but I was curious what the usual approach you would take on this would be.

deploy/scripts/src/mobile/sts/utils/crypto.ts
jamielutton-gds, line 1: The reason I've created this file which "recreates" some functions that exist in the common utils is because I noted that the k6 crypto API was deprecated (see [here](https://grafana.com/docs/k6/latest/javascript-api/k6-experimental/webcrypto/)), so I attempted to recreate what was in the common crypto functions that I needed in the mobile utils.

deploy/scripts/src/mobile/sts/utils/assertions.ts
```diff
@@ -0,0 +1,32 @@
...
+            const url = new URL(res.headers.Location)
+            return requiredQueryParams.every(param => url.searchParams.has(param))
+          }
+        }
+      : {})
+  }
+}
+
+// TODO: Ask if these types of assertions should be kept or not - it is implied by the STS specification that if the status code is 200, then these fields should already be present, so perhaps checking their presence is redundant for a performance test
+export function validateGenerateClientAttestationResponse(res: Response): boolean {
```
jamielutton-gds, line 22: Question: Are assertions on the response body useful whenever the status code implies that the body will have a particular format? For example, in this case, whenever the /generate-client-attestation endpoint returns a 200 status code, it is implied by the specification that it will then include a `client_attestation` field in the response body. So, does this mean that it is redundant to be doing this validation on the presence of `client_assertion` in the response body?

deploy/scripts/src/mobile/sts/testSteps/backend.ts
```diff
@@ -0,0 +1,159 @@
...
+      isStatusCode302,
+      ...validateRedirect(`${config.orchestrationBaseUrl}/authorize`, [
+        'client_id',
+        'scope',
+        'response_type',
+        'request'
+      ])
+    }
+  )
+  return res.headers.Location
```
abhinaysrivastavaperf, line 36: Is the returned redirect Location being used anywhere in the script?

deploy/scripts/src/mobile/v2-sts-get-service-access-token.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export async function getServiceAccessToken(): Promise<void> {
+  const keyPair = await generateKey()
+  const publicKeyJwk = await crypto.subtle.exportKey('jwk', keyPair.publicKey)
+
+  const codeVerifier = crypto.randomUUID()
+  const codeChallenge = await generateCodeChallenge(codeVerifier)
+
+  iterationsStarted.add(1)
+  const orchestrationAuthorizeUrl = getAuthorize(codeChallenge)
```
abhinaysrivastavaperf, line 62: Suggest using a random sleep time between the steps to simulate real world scenario

deploy/scripts/src/mobile/sts/utils/assertions.ts
abhinaysrivastavaperf: In cases like this where the status code of 200 means that the client_attestation will definitely be returned as per the API spec, we can skip the assertion on the response body.
Usually we use a combination of response status code and page content check which is more suitable for pages where an unexpected response could be returned with a 200 status code.

deploy/scripts/src/mobile/sts/utils/crypto.ts
```diff
@@ -0,0 +1,40 @@
+import { algParamMap, JwtAlgorithm } from '../../../common/utils/authentication/jwt'
+import { b64encode } from 'k6/encoding'
+
+// I'm aware that there are existing cryptographic functions in the common utils
```
abhinaysrivastavaperf, line 4: Yeah the common utility needs to be updated. We can validate this script and once this is working fine, we can replace the functions in the common utility and the associated typedoc documentation

abhinaysrivastavaperf (changes requested):

deploy/scripts/src/mobile/sts/utils/assertions.ts
abhinaysrivastavaperf: As highlighted in one of the other comments, if a response if 200 OK definitely means that correct fields are present in the response then pageContentCheck can be skipped.

deploy/scripts/src/mobile/sts/testSteps/backend.ts
jamielutton-gds: Yep - the redirect Location is returned [here](https://github.com/govuk-one-login/performance-testing/pull/1247/files#diff-a4f474262b878343aaf4a9385d71c33b5f3507c734df9ca00b8eb483ecdf2f5bR62) as `orchestrationAuthorizeUrl`, and then used [here](https://github.com/govuk-one-login/performance-testing/pull/1247/files#diff-a4f474262b878343aaf4a9385d71c33b5f3507c734df9ca00b8eb483ecdf2f5bR64).

deploy/scripts/src/mobile/sts/utils/assertions.ts
jamielutton-gds: Have removed any unnecessary assertions

deploy/scripts/src/mobile/v2-sts-get-service-access-token.ts
jamielutton-gds: There's only one point in the flow where we would expect a wait between the API calls due to awaiting for user interaction when they log in via the Auth web screens (email, password, MFA) - I've added a sleep for this between 1-2 seconds (even though in a real scenario it will take longer for the user to do that). Everything else is an automatic sequence of calls made by the app.

abhinaysrivastavaperf (approved):
Added the load profile, parameters for parameter store and the values have been added in the parameter store for both Dev and Build.

PR #1268, opened 2025-04-28
QA-967 - Setup DNS Stacks to allow Public DNS

andyloughran:
### What?
Adds two stacks

1) dns stack
2) acm certificates stack

These both need deploying manually as per the infrastructure deployment process.
See the directions on the ticket for pre-requisites that need completing before this PR can be merged.

#### Changes:
- Adds DNS stack with a hosted zone for each account.
- Adds ACM stack which creates a wildcard certificate for each account.
- Modifies the existing template to references the two previous stacks to allow the rp-stub to run on a public dns domain.

---

### Why?
To allow the rp-stub to be accessible via public DNS, so in the perf account, the API Gateway VPCe can be permanently configured to enabled.


dns/acm-certificate.yaml
```diff
@@ -0,0 +1,77 @@
+AWSTemplateFormatVersion: "2010-09-09"
+
+Description: >-
+  Creates the necessary components to manage DNS for the Performance Test Setup.
+
+Mappings:
+  PlatformConfiguration:
+    build:
+      DNSSUFFIX: perf.build.account.gov.uk
```
gavinmontgomery-gds, line 9: Nit picking - why we calling it `DNSSUFFIX` if it is just the Domain name?

koa-stub/template.yaml
```diff
@@ -56,6 +60,15 @@ Resources:
     Properties:
       StageName: prod
       BinaryMediaTypes: ["*/*"]
+      Domain:
+        CertificateArn: "{{ resolve:ssm:PrimaryZoneWildcardCertificateARN }}"
+        DomainName: !If
+          - IsStaging
+          - "rp-stub.perf.account.gov.uk"
```
andyloughran, line 67:
```suggestion
          - "rp-stub.perf.staging.account.gov.uk"
```

dns/acm-certificate.yaml
andyloughran: Its the SUFFIX for the domains within the account, rather than a specific domain by itself.  The certificate is *.

abhinaysrivastavaperf (approved):

PR #1272, opened 2025-04-29
[QA-972]: Adding regression Load profiles for Address and Fraud

NarasimhaYanamala:
### What?
Load profile for PERF006 - regression testing

#### Changes:
-  created  a load profile function for createI3RegressionScenario
       Ramp up  to rampupNFR and Steady State 5 minutes
  

---

### Why?
To Perform PERF006-Regression tests

---

### Related:

### DIFF
--- a/deploy/scripts/src/common/utils/config/load-profiles.ts
+++ b/deploy/scripts/src/common/utils/config/load-profiles.ts
@@ -104,8 +104,7 @@ export enum LoadProfile {
   spikeNFRSignInL2,
   spikeI2LowTraffic,
   spikeI2HighTraffic,
-  steadyStateOnly,
-  regression
+  steadyStateOnly
 }
 function createStages(type: LoadProfile, target: number): Stage[] {
   switch (type) {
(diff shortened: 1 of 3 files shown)


abhinaysrivastavaperf:
This was not the requirement. The ask was to run the regression test for Address and Fraud. For that the best way is to create separate regression load profiles for Address and Fraud as each one of these will have different ramp up duration as per PERF008

deploy/scripts/src/common/utils/config/load-profiles.ts
```diff
@@ -375,3 +375,27 @@ export function createI3SpikeOLHScenario(
...
+  iterationDuration: number = 5,
+  rampUpNFR: number
+): ScenarioList {
+  const list: ScenarioList = {}
+  const preAllocatedVUs = Math.round((target * iterationDuration) / 2)
+  const maxVUs = target * iterationDuration
+  list[exec] = {
+    executor: 'ramping-arrival-rate',
+    startRate: 1,
+    timeUnit: '1s',
```
abhinaysrivastavaperf, line 391: As per PERF008, the ramp up rate is 1 every 10 seconds.

abhinaysrivastavaperf (changes requested):

NarasimhaYanamala:
> This was not the requirement. The ask was to run the regression test for Address and Fraud. For that the best way is to create separate regression load profiles for Address and Fraud as each one of these will have different ramp up duration as per PERF008

ok, updated the accordingly.

abhinaysrivastavaperf (approved):

abhinaysrivastavaperf (approved):

PR #1274, opened 2025-04-29
[QA-970] Updating the soak test volumes for Auth & Orch.

maddiemorris:
### What?
Updates the soak test volumes for auth & orch `signIn` and `signUp` to 3 and 2 iterations per second respectively. 

#### Changes:
Updates the `perf006Iteration3SoakTest` volumes for `authentication/test.ts` `signIn` and `signUp` to 3 and 2 iterations per second respectively. 

---

### Why?
To run an auth & orch soak test at the above target volumes.

### DIFF
--- a/deploy/scripts/src/authentication/test.ts
+++ b/deploy/scripts/src/authentication/test.ts
@@ -195,7 +195,7 @@ const profiles: ProfileList = {
       executor: 'ramping-arrival-rate',
       startRate: 1,
       timeUnit: '10s',
-      preAllocatedVUs: 30,
+      preAllocatedVUs: 33,
       maxVUs: 66,
       stages: [
         { target: 20, duration: '21s' },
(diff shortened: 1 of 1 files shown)


deploy/scripts/src/authentication/test.ts
```diff
@@ -196,10 +196,10 @@ const profiles: ProfileList = {
       startRate: 1,
       timeUnit: '10s',
       preAllocatedVUs: 30,
```
abhinaysrivastavaperf, line 198: Just being nitpicky -Should the preAllocatedVUs count change as well?

abhinaysrivastavaperf (changes requested):

deploy/scripts/src/authentication/test.ts
maddiemorris: @abhinaysrivastavaperf should `preAllocatedVUs` be half of `maxVUs`?

PR #1316, opened 2025-05-13
[QA-1060]: Update I4 PeakTest SignUp and Sign In Load Profile to adjust the VU counts

umairfarooquigds:
### What?
Update I4 PeakTest SignUp Load Profile to adjust the VU counts

#### Changes:
- Target for pre Allocated VUs should not be 10x hence the change is `const preAllocatedVUs = Math.round(((target / 10) * iterationDuration) / 2)`
- Target for Max VUs should not be 10x hence the change is `const maxVUs = Math.round((target / 10) * iterationDuration)`

---

### Why?
To initiate the right number of pre-allocated and max virtual users during the peak test for iteration 4.

### DIFF
--- a/deploy/scripts/src/common/utils/config/load-profiles.ts
+++ b/deploy/scripts/src/common/utils/config/load-profiles.ts
@@ -351,8 +351,8 @@ export function createI4PeakTestSignUpScenario(
   rampUpDuration: number
 ): ScenarioList {
   const list: ScenarioList = {}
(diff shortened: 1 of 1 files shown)


abhinaysrivastavaperf (changes requested):
Can you please update `createI4PeakTestSignInScenario` function as the maxVUs calculation is incorrect?

abhinaysrivastavaperf (approved):

PR #1319, opened 2025-05-14
QA-1064: I4 Peak Load profile update to Orange Address script

vineeth-kunju:
### What?
To add peak load profile to Orange Address script

#### Changes:
Added `perf006Iteration4PeakTest` profile for address & addressME scenarios to achieve the below peak targets
`address: 10 iters/sec`
`addressME: 37 iters/sec` - the rampup starts after the initial 101 seconds wait

---

### Why?
To test Perf006 Iters4 peak test

---

### DIFF
--- a/deploy/scripts/src/cri-orange/address.ts
+++ b/deploy/scripts/src/cri-orange/address.ts
@@ -146,6 +146,33 @@ const profiles: ProfileList = {
   perf006Iteration3SpikeTest: {
     ...createI3SpikeSignUpScenario('address', 100, 15, 101),
     ...createI3SpikeSignUpScenario('addressME', 390, 15, 391)
+  },
+  perf006Iteration4PeakTest: {
+    address: {
+      executor: 'ramping-arrival-rate',
+      startRate: 1,
+      timeUnit: '10s',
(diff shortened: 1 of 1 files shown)


deploy/scripts/src/cri-orange/address.ts
```diff
@@ -146,6 +146,33 @@ const profiles: ProfileList = {
   perf006Iteration3SpikeTest: {
     ...createI3SpikeSignUpScenario('address', 100, 15, 101),
     ...createI3SpikeSignUpScenario('addressME', 390, 15, 391)
+  },
+  perf006Iteration4PeakTest: {
+    address: {
```
abhinaysrivastavaperf, line 151: You can just use the `createI4PeakTestSignUpScenario` function for the normal address scenario

abhinaysrivastavaperf (changes requested):

deploy/scripts/src/cri-orange/address.ts
vineeth-kunju: Hope this works

abhinaysrivastavaperf (approved):

PR #1333, opened 2025-05-16
QA-1063 IPVcore script - use Orch stub + devperf

richie-co:
### What?
Update IPVCore script to use Orch stub with devperf 

#### Changes:
- update environment constants to include PERF 

---

### Why?
To enable using Orch stub (Perf) along with the dev - perf environment and support regression testing

---

### DIFF
--- a/deploy/scripts/src/ipv-core/test.ts
+++ b/deploy/scripts/src/ipv-core/test.ts
@@ -693,7 +693,6 @@ export function identity(stubOnly: boolean = false): void {
     )
   })
   iterationsCompleted.add(1)
-  console.log(userId, testEmail)
 }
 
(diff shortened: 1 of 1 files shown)


deploy/scripts/src/ipv-core/test.ts
```diff
@@ -693,6 +693,7 @@ export function identity(stubOnly: boolean = false): void {
     )
   })
   iterationsCompleted.add(1)
+  console.log(userId, testEmail)
```
abhinaysrivastavaperf, line 696: Is this to create test data?

abhinaysrivastavaperf (changes requested):

PR #1360, opened 2025-05-22
[QA-1109] TiCF Script changes

maddiemorris:
### What?
Creates one single function, `ticf()`, in the `fraud/ticf.ts` script that executes all four scenarios (`signInsuccess`, `signUpSuccess`, `identityProvingSuccess`, and `identityReuseSuccess`) inside it.

#### Changes:
- changes the loadProfile and groupMap for the new `ticf()` function
- creates a new single function, `ticf()`, which calls the `signUpSuccess`, `signInsuccess`, `identityProvingSuccess`, and `identityReuseSuccess` functions in order, maintaining a constant userID between them. 

---

### Why?
To ensure all four scenarios are being executed with the same userID in the correct order, for a successful performance test.

### DIFF
--- a/deploy/scripts/src/fraud/ticf.ts
+++ b/deploy/scripts/src/fraud/ticf.ts
@@ -25,19 +25,18 @@ import { timeGroup } from '../common/utils/request/timing'
 
 const profiles: ProfileList = {
   smoke: {
-    ...createScenario('signInSuccess', LoadProfile.smoke),
-    ...createScenario('signUpSuccess', LoadProfile.smoke),
-    ...createScenario('identityProvingSuccess', LoadProfile.smoke),
-    ...createScenario('identityReuseSuccess', LoadProfile.smoke)
+    ...createScenario('ticf', LoadProfile.smoke)
   }
 }
 
 const loadProfile = selectProfile(profiles)
 const groupMap = {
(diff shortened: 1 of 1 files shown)


abhinaysrivastavaperf (changes requested):

deploy/scripts/src/fraud/ticf.ts
```diff
@@ -68,87 +67,78 @@ const awsConfig = new AWSConfig({
...
 const sqs = new SQSClient(awsConfig)
 
-export function signInSuccess(): void {
-  const groups = groupMap.signInApiCall
-  const userID = `urn:fdc:gov.uk:2022:${uuidv4()}`
+export function signUpSuccess(userID: string): void {
+  const groups = groupMap.ticf
+  const timestamp = new Date().toISOString().slice(0, 19).replace(/[-:]/g, '') // YYMMDDTHHmmss
+  const testID = `perfTestID${timestamp}`
   const emailID = `perfEmail${uuidv4()}@digital.cabinet-office.gov.uk`
```
abhinaysrivastavaperf, line 74: The email ID should be the same as well. This can also parsed as a parameter.

abhinaysrivastavaperf (changes requested):

abhinaysrivastavaperf (approved):

PR #1379, opened 2025-05-29
[QA-1130] Creating dynamic eventID values for TxMA events in the TiCF script.

maddiemorris:
### What?
created dynamic & unique `eventID` values for each TxMA event sent in the `ticf.ts` script.

#### Changes:
- Adds the `eventID` value back into the TxMA Request Generator instead of passing it as a parameter, to ensure each value of `eventID` is unique for each TxMA event sent. 

---

### Why?
To create a fully representative/real-world performance scenario.

### DIFF
--- a/deploy/scripts/src/common/requestGenerator/txmaReqGen.ts
+++ b/deploy/scripts/src/common/requestGenerator/txmaReqGen.ts
@@ -145,10 +145,10 @@ export function generateDcmawAbortWeb(userID: string, journeyID: string, emailID
   }
 }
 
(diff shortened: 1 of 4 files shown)


abhinaysrivastavaperf:
Should we be updating the TxMA scripts where the common functions are used?

abhinaysrivastavaperf (changes requested):
Please check TxMA script where we are using some of these common events

abhinaysrivastavaperf (approved):

PR #1394, opened 2025-06-02
DCMAW-13276: Add reauthentication script for STS

jamielutton-gds:
### What?

- Adds `reauthentication` scenario for STS test script. 

#### Changes:

- Adds `reauthentication` scenario for STS test script which is the same as the `authentication` scenario except for the addition of the persistent ID in the `govuk_signin_session_id` query parameter in the authorize request. 
- Adds test data files for reauthentication scenario and adds documentation to README on how to generate new sets of test data when old one is expired.
  - Bash scripts for generating the test data in dev and build are included. 

---

### Why?

To facilitate load testing of reauthentication scenario for STS, which is one of the Wallet user journeys. 

---

### Related:


deploy/scripts/src/mobile/sts.ts
```diff
@@ -0,0 +1,261 @@
+import {
```
jamielutton-gds, line 1: The SonarCloud quality gate fails due to duplication - this could be because I have added code to this branch as in the branch for [this PR](https://github.com/govuk-one-login/performance-testing/pull/1366) - I will wait for the other PR to be merged and then check if duplication is still an issue before resolving.

abhinaysrivastavaperf (approved):
Script yet to be validated. Updated incorrect PR.

abhinaysrivastavaperf (changes requested):
Will validate script before merging

abhinaysrivastavaperf (approved):

PR #1404, opened 2025-06-06
[QA-1160] Profile change for  Mobile wallet Credential Issuance for BE and STS 

NarasimhaYanamala:
### What?
Profile change for  Mobile wallet Credential Issuance for BE and STS 

#### Changes:
- Mobile Wallet Backend Credential Issuance
   Ramp up 18 sec and Steady state duration is 55 min
- Mobile Wallet STS Credential Issuance
   Ramp up 18 sec and Steady state duration is 55 min
---

### Why?
To perform Peak testing for Wallet

---


abhinaysrivastavaperf (changes requested):
Change the profile name in Backend script to `walletPerfTestBackend`

PR #1458, opened 2025-06-26
QA-1216: Adding Signin Volume for GetContraIndicator API

vineeth-kunju:
### What?
This PR is to add SignIn volume along with the existing SignUp volume for `GetContraIndicator` API

#### Changes:
- Created a new `cimitSignInAPI` scenario to trigger the `GetContraIndicator` API for the desired SignIn Volume
- Renamed the existing `cimitAPIs` scenario to `cimitSignUpAPIs` and applied the change to all the existing profiles. 
- Added the new to `cimitSignInAPI` scenario to `perf006Iteration5PeakTest` profile to generate 65 iterations per second

Note: Other profiles like `lowvolume`, `load`, etc can be updated based on the success of this approach

---

### Why?
To accomodate the `GetContraIndicator` api calls generated during SignIn 

---

### DIFF
--- a/deploy/scripts/src/cimit/test.ts
+++ b/deploy/scripts/src/cimit/test.ts
@@ -5,6 +5,7 @@ import {
   createScenario,
   LoadProfile,
   createI4PeakTestSignUpScenario,
+  createI4PeakTestSignInScenario,
   createI3SpikeSignUpScenario
 } from '../common/utils/config/load-profiles'
 import http from 'k6/http'
@@ -22,22 +23,24 @@ import { uuidv4 } from '../common/utils/jslib'
 
 const profiles: ProfileList = {
   smoke: {
-    ...createScenario('cimitAPIs', LoadProfile.smoke)
+    ...createScenario('cimitSignUpAPIs', LoadProfile.smoke),
+    ...createScenario('cimitSignInAPI', LoadProfile.smoke)
   },
   lowVolume: {
-    ...createScenario('cimitAPIs', LoadProfile.short, 30, 5)
+    ...createScenario('cimitSignUpAPIs', LoadProfile.short, 30, 5)
   },
   load: {
-    ...createScenario('cimitAPIs', LoadProfile.full, 400, 5)
+    ...createScenario('cimitSignUpAPIs', LoadProfile.full, 400, 5)
   },
   perf006Iteration4PeakTest: {
-    ...createI4PeakTestSignUpScenario('cimitAPIs', 1880, 19, 471)
+    ...createI4PeakTestSignUpScenario('cimitSignUpAPIs', 1880, 19, 471)
   },
   perf006Iteration4SpikeTest: {
-    ...createI3SpikeSignUpScenario('cimitAPIs', 4520, 19, 1131)
+    ...createI3SpikeSignUpScenario('cimitSignUpAPIs', 4520, 19, 1131)
   },
   perf006Iteration5PeakTest: {
-    ...createI4PeakTestSignUpScenario('cimitAPIs', 2280, 19, 571)
+    ...createI4PeakTestSignUpScenario('cimitSignUpAPIs', 2280, 19, 571),
+    ...createI4PeakTestSignInScenario('cimitSignInAPI', 65, 6, 30)
   }
 }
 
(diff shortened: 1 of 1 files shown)


deploy/scripts/src/cimit/test.ts
```diff
@@ -120,3 +123,24 @@ export async function cimitAPIs(): Promise<void> {
...
+    headers: {
+      'govuk-signin-journey-id': uuidv4(),
+      'ip-address': '1.2.3.4'
+    }
+  }
+
+  iterationsStarted.add(1)
+
+  // B02_CIMIT_03_GetContraIndicatorCredentials
+  timeGroup(groups[1], () => http.get(env.envURL + `/v1/contra-indicators?user_id=${subjectID}`, params), {
```
abhinaysrivastavaperf, line 141: The user ID needs to be in the DB through a PutContraIndicator or a PostMitigation call before GetContraIndicator is called.
With this code, you are searching for a random new user which doesn't exist in the system.

abhinaysrivastavaperf (changes requested):

deploy/scripts/src/cimit/test.ts
abhinaysrivastavaperf: Where are you using the data from a CSV in this PR?

abhinaysrivastavaperf (changes requested):

deploy/scripts/src/cimit/test.ts
```diff
@@ -120,3 +132,69 @@ export async function cimitAPIs(): Promise<void> {
...
+    isStatusCode200,
+    ...pageContentCheck('vc')
+  })
+  iterationsCompleted.add(1)
+}
+
+export async function generateCIs(): Promise<void> {
```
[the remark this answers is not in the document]
vineeth-kunju: Removed

deploy/scripts/src/cimit/test.ts
```diff
@@ -120,3 +132,69 @@ export async function cimitAPIs(): Promise<void> {
     ...pageContentCheck('vc')
   })
 }
+
+export async function cimitSignInAPI(): Promise<void> {
+  const groups = groupMap.cimitAPIs
```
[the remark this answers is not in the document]
vineeth-kunju: Updated, please review

deploy/scripts/src/cimit/test.ts
```diff
@@ -22,32 +24,44 @@ import { uuidv4 } from '../common/utils/jslib'
...
 const loadProfile = selectProfile(profiles)
 const groupMap = {
-  cimitAPIs: [
+  cimitSignUpAPIs: [
     'B01_CIMIT_01_PutContraIndicator',
     'B03_CIMIT_02_PostMitigations',
     'B02_CIMIT_03_GetContraIndicatorCredentials'
-  ]
+  ],
+  cimitSignInAPIs: ['B02_CIMIT_03_GetContraIndicatorCredentials']
```
abhinaysrivastavaperf, line 64: If the transaction name of sign in is same in both the functions, how will we differentiate it in the `durations.csv` for the two functions?

deploy/scripts/src/cimit/test.ts
```diff
@@ -115,8 +144,29 @@ export async function cimitAPIs(): Promise<void> {
...
-  timeGroup(groups[1], () => http.get(env.envURL + `/v1/contra-indicators?user_id=${subjectID}`, params), {
+  timeGroup(groups[2], () => http.get(env.envURL + `/v1/contra-indicators?user_id=${subjectID}`, params), {
     isStatusCode200,
     ...pageContentCheck('vc')
   })
+  console.log(subjectID)
+}
+
+export async function cimitSignInAPI(): Promise<void> {
+  const retrieveData = csvData[Math.floor(Math.random() * csvData.length)]
```
abhinaysrivastavaperf, line 155: Using random here would mean that multiple VUs can use the same userID which is not a representative scenario for production.
This should be unique to each user

abhinaysrivastavaperf, line 155:
```suggestion
  const retrieveData = csvData[execution.vu.idInTest - 1]
```

abhinaysrivastavaperf (changes requested):

deploy/scripts/src/cimit/test.ts
vineeth-kunju: I had this in mind, since the Get calls are same, i kept it as it is. Also, the intention was to include additional SignIn volumes. 
Updated the transaction names and changed the order properly

deploy/scripts/src/cimit/test.ts
vineeth-kunju: Changed according to your comment. Unique for each user, i need to explore and understand in mechanism in K6. thanks

PR #1593, opened 2025-10-14
QA-1364: PR to include Dynamic(EPOCH) journID for TXMA

vineeth-kunju:
### What?
This PR is to replace the current static journeyID in the TXMA test script to a dynamic value(EPOCH timestamp), which would help in debugging

#### Changes:
- Replaced the current static value of journeyID with Math.floor(Date.now() / 1000).toString() in the setup function, so that the value remains same for each test
- journeyID is returned from the setup function and used in the sendRegularEventWithEnrichment function

### Why?
A unique journeyID for each test would help in filtering, debugging etc.

---

### DIFF
--- a/deploy/scripts/src/txma/clientEnrichment.ts
+++ b/deploy/scripts/src/txma/clientEnrichment.ts
@@ -118,14 +118,14 @@ const awsConfig = new AWSConfig({
 
 const sqs = new SQSClient(awsConfig)
 
-export function setup(): string {
+export function setup() {
   describeProfile(loadProfile)
   const timestamp = new Date().toISOString().slice(0, 19).replace(/[-:]/g, '') // YYMMDDTHHmmss
   const testID = `perfTestID${timestamp}`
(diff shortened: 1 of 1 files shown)


abhinaysrivastavaperf (changes requested):

deploy/scripts/src/txma/clientEnrichment.ts
```diff
@@ -136,15 +136,14 @@ export function setup(): string {
   console.log('Sending primer event 2')
   sqs.sendMessage(env.sqs_queue, authReqParsedPayloadEnrichment)
   console.log('Primer event 2 sent')
-  return authCreateAccPayload
+  return { authCreateAccPayload: authCreateAccPayload, journeyID: journeyID }
```
abhinaysrivastavaperf: Have we tested if this returns the correct values?

abhinaysrivastavaperf (changes requested):

PR #1876, opened 2026-04-21
QA-1688: IPVCore Expired DL TestData for IDReuse scenario

vineeth-kunju:
### What?
This PR adds test data for validating that an expired driving licence can be used for IdReuse scenario within 3 months grace period

#### Changes:

- Updated drivingLicenceData.ts — changed driving permit expiryDate from 2032-02-02 to 2026-03-21 to simulate an expired licence within the 3-month grace period
- 5k rows of test data generated locally 
- Appended new user records to idReuseTestData.csv generated using the expired DL test data

---

### Why?
To conduct IPVCore regression test against I8 volume and to prove the logic for validating expired DL will not have any performance impact

---


NarasimhaYanamala (approved):
ok

PR #1887, opened 2026-05-01
[QA-1688]: TxMA - Update event payload as per latest guidance

abhinaysrivastavaperf:
### What?
TxMA - Update event payload as per latest guidance

#### Changes:
- Add client_id field as specified in the event catalogue.

---

### Why?
To keep the performance test payload in line with the latest guidance.


marianneolabode (approved):
Approved

PR #1919, opened 2026-06-02
QA-1705: Add high volume SPOT load profile for Cribl log ingestion validation

vineeth-kunju:
### What?
This PR is to add a new load profile `criblHighVolumeTest` to spot test script to support a high volume test, validating that the Cribl instance for the new logging solution can sustain high volumes of log ingestion in an hour.

#### Changes:
- **`load-profiles.ts`**: included a new helper `createHighVolumeSignInScenario` that  wraps `createPerfTestScenario` with a 1-hour hold duration
- **`spot.ts`**: Added `criblHighVolumeTest` profile using the new helper — ramps up to 3000 req/s in 22.5 minutes then run for 60 minutes

---

### Why?
To support Cribl instance implementation

---


abhinaysrivastavaperf (approved):
This is fine on a temporary basis. The overall load profile config needs an overhaul to avoid so much duplication.

PR #1930, opened 2026-06-12
[QA-1718]:Add I10 peaktest profile for passkey sceanrios

marianneolabode:
### What?
Add I10 peak test profile for passkey scenarios and other 

#### Changes:
-Update to 
-- Added sleepBetween(1, 3) before sign-out in landingPage function for realistic user simulation
- add perf006Iteration10PeakTest scenarios with new load parameters:
  - OLH scenarios: 12 iterations/min (0.2 iter/sec each)
  - landingPage: 1038 iterations/sec
  - Updated iteration durations based on performance requirements". ==> The total expected duration is calculated by multiplying the number of steps by the worst‑case response time (we anticipate 3 sec)and adding the average think time—two seconds—for each of the gaps between steps(random between 1 and 3 sec =2).

---

### Why?
To Test passkey at Iteration 10 peak load


marianneolabode:
Fixed lint issue

deploy/scripts/src/olh/test.ts
```diff
@@ -415,7 +415,10 @@ const profiles: ProfileList = {
     ...createOLHPeakTestScenario('changePassword', 6, 21, 1),
     ...createOLHPeakTestScenario('changePhone', 6, 24, 1),
     ...createOLHPeakTestScenario('deleteAccount', 6, 18, 1),
-    ...createI4PeakTestSignInScenario('landingPage', 12, 9, 6)
+    ...createI4PeakTestSignInScenario('landingPage', 12, 9, 6),
+    // Passkey scenarios (accounting for their sleep patterns):
```
abhinaysrivastavaperf: We do not need the comment here.

deploy/scripts/src/olh/test.ts
```diff
@@ -415,7 +415,10 @@ const profiles: ProfileList = {
...
+    ...createOLHPeakTestScenario('setUpPasskey', 6, 28, 1), // ~5 sleepBetween calls + AMC complexity
```
abhinaysrivastavaperf: Not sure if I understand what you mean by AMC complexity here.
Set up passkey has 7 steps so the max iteration duration can be 21 seconds + think time.
Same for remove passkey scenario.

abhinaysrivastavaperf (changes requested):
Please add an Iteration 10 peak test profile with the volumes that you shared in the table on slack.

deploy/scripts/src/olh/test.ts
marianneolabode: Other scenarios in the test with 7–8 steps and the same think times as the passkey scenarios don't have iteration durations anywhere near 28 seconds.
If I apply the 21-second response time plus think time, the expected iteration duration would be approximately:

21 + (6 × 2) = 33 seconds (using average think time)
21 + (6 × 3) = 39 seconds (using max think time)
However, based on the pattern observed across other scenarios, the actual calculation being applied appears to be:
(3 × number of steps )+ minimum think time 

landing page = (3*3) +3 = 12 
delete account  (6*3 )+ 0 = 18 - missing thing time
change PWD (7*3 )+ 0 = 21 - missing thing time
change email (8*3 )+ 0 = 24- missing thing time
change phone (8*3 )+ 0 = 24.- missing thing time
==>Passkey scenarios (7*3)+ 7 = 28. though there are n-1 = 6 think times and not n = 7

Should i go for 27, 28 , 33, or 39 sec

abhinaysrivastavaperf (approved):

abhinaysrivastavaperf:
Just a minor comment - Place the Iteration 10 profile at the last so that it is easy to track

deploy/scripts/src/olh/test.ts
```diff
@@ -417,6 +417,16 @@ const profiles: ProfileList = {
     ...createOLHPeakTestScenario('deleteAccount', 6, 18, 1),
     ...createI4PeakTestSignInScenario('landingPage', 12, 9, 6)
   },
+
+  perf006Iteration10PeakTest: {
+    ...createOLHPeakTestScenario('changeEmail', 12, 38, 1),
+    ...createOLHPeakTestScenario('changePassword', 12, 33, 1),
+    ...createOLHPeakTestScenario('changePhone', 12, 38, 1),
+    ...createOLHPeakTestScenario('deleteAccount', 12, 28, 1),
+    ...createI4PeakTestSignInScenario('landingPage', 1038, 13, 6),
```
marianneolabode, line 426:
```suggestion
    ...createI4PeakTestSignInScenario('landingPage', 18, 13, 6),
```

marianneolabode (commented):
the target iteration should be 18 per sec

PR #1944, opened 2026-06-16
[QA-1720]:Correct the Iteration 10 peak‑test landing‑page target iteration value.

marianneolabode:
### What?
Correct the value of the landing page.

#### Changes:
- Detail individual changes in bullet points

Correct the value of the landing page function target from 1038  per to 18  so it runs at 1038 journeys per sec/18 journeys per min.

### Why?
To run the I10 peak test at the correct journeys rate for landing page 

---


abhinaysrivastavaperf (changes requested):
The volume targets look alright but as mentioned in previous PR review, the latest profile needs to be kept at the last for tracking. Can you move this please?

marianneolabode:
> The volume targets look alright but as mentioned in previous PR review, the latest profile needs to be kept at the last for tracking. Can you move this please?

Done

abhinaysrivastavaperf (approved):

PR #1948, opened 2026-06-19
[QA-1724]: Load Profiles - Align the steady state durations of Sign Up and Sign in scenarios

abhinaysrivastavaperf:
### What?
Load Profiles - Align the steady state durations of Sign Up and Sign in scenarios

#### Changes:
- Currently for all our peak tests, the steady state timeline for Sign Up and Sign in is different as the ramp up durations are different. This results in the steady state load not being aligned for the entire duration.
- We are updating the createPerfTestScenario() function to add a phaseDelay parameter. 
- For scenarios with shorter ramp up duration we should add a phaseDelay value which would be the difference between the longer ramp up duration and the shorter ramp up duration.
- With this PR, we are also updating the volume targets for Authentication Iteration 10 testing. Attaching a screenshot showing evidence that the change has been validated locally and the steady state durations are now aligned

image

---

### Why?
To align the steady state durations of Sign Up and Sign in scenarios

### DIFF
--- a/deploy/scripts/src/authentication/test.ts
+++ b/deploy/scripts/src/authentication/test.ts
@@ -341,8 +341,8 @@ const profiles: ProfileList = {
   },
   perf006Iteration10PeakTest: {
     ...createI4PeakTestSignUpScenario('signUp', 75, 43, 751),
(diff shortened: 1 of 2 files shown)


NarasimhaYanamala (approved):
ok

PR #1974, opened 2026-07-09
[QA-1740]: Transit Gateway: Remove content validation from client lambda call

abhinaysrivastavaperf:
### What?
Transit Gateway: Remove content validation from `client` lambda call

#### Changes:
- Remove the page content validation from the `client` lambda call.

---

### Why?
The lambda invocation just returns a 200 OK response with a blank response body. We have cross checked the application logs and the lambda is working as expected.


NarasimhaYanamala (approved):
ok

PR #1991, opened 2026-07-14
[QA-1774]:Orange Ex KBV - Add I10 peak load profile

marianneolabode:
### What?
Add Iteration 10 peak load Test profile:

#### Changes:
Add Iteration 10 peak load Test profile:
-KBV at 2 Journeys per sec.

---

### Why?
To test the i10 peak test profiles

---

### DIFF
--- a/deploy/scripts/src/cri-orange/kbv.ts
+++ b/deploy/scripts/src/cri-orange/kbv.ts
@@ -111,6 +111,9 @@ const profiles: ProfileList = {
   },
   perf006Iteration9StressTest: {
     ...createStressTestSignUpScenario('kbv', 63, 12, 64)
+  },
+  perf006Iteration10PeakTest: {
+    ...createI4PeakTestSignUpScenario('kbv', 20, 12, 21)
   }
 }
 


deploy/scripts/src/cri-orange/kbv.ts
```diff
@@ -111,6 +111,9 @@ const profiles: ProfileList = {
   },
   perf006Iteration9StressTest: {
     ...createStressTestSignUpScenario('kbv', 63, 12, 64)
+  },
+  perf006Iteration10PeakTest: {
```
abhinaysrivastavaperf, line 115: We do not need to create a new profile for KBV as the same volume was tested in Iteration 8. Just reuse the `perf006Iteration8PeakTest` profile for this test and close this PR please.

abhinaysrivastavaperf (changes requested):

deploy/scripts/src/cri-orange/kbv.ts
marianneolabode: OK.

marianneolabode:
The existing I8 profileas will be used as it is the same as the I10 profile. hence closing this PR

PR #1992, opened 2026-07-14
[QA-1778]: Mobile ID Check Async - Add Iteration 10 load profile

abhinaysrivastavaperf:
### What?
Mobile ID Check Async - Add Iteration 10 load profile

#### Changes:
Add Iteration 10 load profile for ID Check Async scenarios with below test targets
- ID Check scenario - 18 journeys per second.
- Sign in scenario - 267 journeys per second

---

### Why?
To run Iteration 10 peak test for ID Check async scripts.


NarasimhaYanamala (approved):
ok

PR #1993, opened 2026-07-15
[QA-1776]: Mobile App V1 - Fix the issue with redirect step and add Iteration 10 load profile.

abhinaysrivastavaperf:
### What?
Mobile App V1 - Fix the issue with redirect step and add Iteration 10 load profile.

#### Changes:
- Replaced the `/redirect` call to be a FE call instead of the BE call. Added the extraction of `authorizationCode` and `redirectUrl` from the response of the FE call.
- Added the Iteration 10 peak test load profile.
- The script has been validated locally

image

---

### Why?
To conduct Iteration 10 peak test for Mobile App v1 journey.


NarasimhaYanamala (approved):
ok

sandymay-dd (approved):

PR #2005, opened 2026-07-23
DCMAW-21668: Update Status List Api performance tests using correct integration method

dothomson:
# Status List Performance Tests

The Status list are moving from a private api gateway secured at a network level using a vpc endpoint to a public Status Api secured with a sigv4 header. This is the same access pattern currently used for the "proxy" api allowing ther performance tests to be tested outside of the vpc.

The Status List have now removed the proxy api in favour of the status api. The old private API will be removed once all integrations hav moved across (there are two other integrations, the example credential issuer from on boarding products and Dvla in the integration environment)

Have validated the script in build locally. This is working fine.
image


abhinaysrivastavaperf (approved):
Look good to me. Has been validated locally for both Dev and Build locally

